Issue #4, opened 2019-03-09
Able to filter gzipped vcf files but not uncompressed vcf files


sameerd:
Hi, 

Thanks for the dumpSTR filtering tool.  It is quite nifty. 

I'm able to read compressed vcf files but not uncompressed vcf files with a few of the latest versions including the current master branch. Please see below. 

```
$ PATH=PATH:/projects/b1049/genetics_programs/gangSTR/bin PYTHONPATH=/projects/b1049/genetics_programs/gangSTR/lib/python3.7/site-packages dumpSTR --vcf SS4009013.vcf --max-call-DP 1000 --out filtered
Traceback (most recent call last):
  File "/projects/b1049/genetics_programs/gangSTR/bin/dumpSTR", line 11, in <module>
    load_entry_point('strtools==1.0.0', 'console_scripts', 'dumpSTR')()
  File "/projects/b1049/genetics_programs/gangSTR/lib/python3.7/site-packages/strtools-1.0.0-py3.7.egg/dumpSTR/dumpSTR.py", line 325, in main
  File "/projects/b1049/genetics_programs/gangSTR/lib/python3.7/site-packages/PyVCF-0.6.8-py3.7-linux-x86_64.egg/vcf/parser.py", line 300, in __init__
    self._parse_metainfo()
  File "/projects/b1049/genetics_programs/gangSTR/lib/python3.7/site-packages/PyVCF-0.6.8-py3.7-linux-x86_64.egg/vcf/parser.py", line 318, in _parse_metainfo
    while line.startswith('##'):
TypeError: startswith first arg must be bytes or a tuple of bytes, not str
```

I have no trouble reading in the file if it is compressed.

gymreklab:
Fixed by #5 thanks to @sameerd

PR #5, opened 2019-03-09
Read vcf files by not specifing the read mode

sameerd:
This is a possible fix for Issue #4 which works for me.


gymreklab:
Thanks for the PR! Yes, I ran into this issue recently as well. I appreciate the fix.

Issue #7, opened 2019-03-17
Import error when running plinkSTR


sameerd:
Hi I get a NameError error when running plinkSTR with no commands because it tries to import its own package called common and a different package called [common already exists in python](
https://pypi.org/project/common). 

```shell
$ plinkSTR
Traceback (most recent call last):
  File "/projects/b1049/genetics_programs/gangSTR/bin/plinkSTR", line 11, in <module>
    load_entry_point('strtools==1.0.0', 'console_scripts', 'plinkSTR')()
  File "/projects/b1049/genetics_programs/gangSTR/lib/python3.7/site-packages/strtools-1.0.0-py3.7.egg/plinkSTR/plinkSTR.py", line 276, in main
NameError: name 'ERROR' is not defined
```

gymreklab:
Addressed in #8

PR #8, opened 2019-03-17
common is a different python package we specify which one is needed

sameerd:
Here is bugfix for package import clash - common #7


gymreklab:
Thanks for the bug fix!

PR #15, opened 2019-04-17
plinkSTR speedups

shubhamsaini:
1. statsmodels.formulas is inherently slow. Using traditional statsmodels.api is 3x faster
2. pdata now required "intercept" column for use with Logit.fit()


shubhamsaini:
plinkSTR updated. moving away from formulas api, to conventional label - features matrices method. compared the two methods, seems to work fine. can now be merged with main branch.

shubhamsaini:
plinkSTR now correctly runs allele based tests, and doesn't crash on missing GT values.
Added Plink style LD-based clumping script for GWAS output from plinkSTR

PR #42, opened 2020-01-28
Adding tests for TR_harmonizer

nmmsv:


nmmsv:
advntr and popstr tests are commented for now because they broke the harmonizer.

>            if vcfrecord.ALT[0] is not None:
>                alt_alleles = []
>                for alt in vcfrecord.ALT:
>                    n_repeats = float(str(alt)[1:-1])
>                    int_repeats = math.floor(n_repeats)
>                    extra = n_repeats - int_repeats
>                   alt_allele = int_repeats*motif
>E                   TypeError: can't multiply sequence by non-int of type float`

PR #43, opened 2020-01-29
Improvements to tr harmonizer test

nmmsv:
Added test for unsupported vcf (snp) to fail gracefully
Added test for incorrect vcf (if we choose advntr for gangstr vcf) to fail gracefully -> we don't fail gracefully and this needs to be fixed


nmmsv:
I think we should add commits for fixing the tests under this PR before merging

gymreklab:
This is breaking Travis: 
```            motif = vcfrecord.INFO["RU"].upper()
>           record_id = vcfrecord.INFO["VID"]
E           KeyError: 'VID'
strtools/utils/tr_harmonizer.py:207: KeyError```

nmmsv:
Yeah that's the error that I get when I run pytest. I think it's because we don't really handle cases that incorrect vcf type is specified, so it tries to find fields that don't exist and throws Key error.
In test I expect value error, and I think that's why it breaks. I think we should add something to tr harmonizer to prevent this key error. I can try working on it if tr harmonizer is not under dev right now.

PR #48, opened 2020-02-07
Vcff test

nmmsv:


nmmsv:
Tests currently break and we need to update mergeSTR to fix it.

PR #49, opened 2020-02-08
Added tests for mergeSTR

nmmsv:


nmmsv:
ALMOST THERE, 99%

nmmsv:
Ok seems like I have another issue. When I try to check warnings, I notice there are a bunch of other warnings for unclosed vcf files. But pyvcf reader does not have a close() function. Because of this Travis fails as the warning is not indexed properly.

PR #72, opened 2020-04-23
Version mod

nmmsv:


nmmsv:
@LiterallyUniqueLogin I added the contributing info to both readme and a text file, and bumped the version since I'm merging into master. After this, I'll create the develop branch and we can use that from now on.

nmmsv:
Hi @LiterallyUniqueLogin , I forgot about this PR :D 
I just merged your latest changed into the branch and I think it should be ready to merge if you think it looks alright.
I basically added a few lines to setup.py that we have to update when we want to bump the version. We also have to run setup.py to actually bump the python file that keeps the version number, but I think that should be alright. I borrowed this method from pybedtools versioning code.
Let me know what you think.

LiterallyUniqueLogin:
Small thing: 'not' is a typo for 'note' in one of the files.

The changes to versioning look good! Thank you, this is cleaner.

I'm not sure a single 'develop' branch is better than having many branches off of master, one for each feature in development. But I'm not sure enough about this to feel comfortable suggesting a change.

Ship it!

LiterallyUniqueLogin:
Whoops, didn't mean to close.

nmmsv:
Thank you for checking this.
Oh I totally meant having one develop branch as latest codebase, and having feature branches from develop and merged back to develop.
The advantage is that we don't have to bump the version for every single minor change, instead we can bump the version every once in a while by having PR from develop to master.
Does it make sense?

LiterallyUniqueLogin:
About the develop branch - that makes sense. My gut is that it will be more hassle to remember to merge from develop to master and to remember to version bump then than to just do it for every feature, given that I don't expect super frequent contributions to this package. But maybe that's incorrect.

Let's do it!

nmmsv:
I agree let's give it a shot haha :D ok I'll merge now

PR #74, opened 2020-05-21
Fixed a dumpSTR test that was failing because it assumed the user had…

LiterallyUniqueLogin:
… no permissions to write files to the root directory '/', but that's not true in some dockered environments


nmmsv:
Looks good to me! thanks for making the update!

PR #75, opened 2020-05-23
Change test setup so that tests can load data from a datadirectory wh…

LiterallyUniqueLogin:
…erever it is (specified by --datadir on the command line) instead of looking in a hardcoded place, so we don't have to include the large data files in the source code distribution, but we can still have people download the data if they want to run the tests (the script `test_trtools.sh` will do that for them). Also, stop throwing extra test files in DUMPDIR, instead throw them in a temporary directory that the OS will automatically clean up.
`
For example, if you move the tests/common directory to tests/common2 and run `pytest .` from the repository root, then you'll get errors because it can't find the test files, but if you run `pytest . --datadir tests/common2` it will work fine because it will find the directory in the location you've specified.

Note that these changes make use of pytest [fixtures](https://docs.pytest.org/en/latest/fixture.html#funcargs). This both for getting access to a canonical location for [temporary directories](https://docs.pytest.org/en/latest/tmpdir.html), as well as [a way](https://docs.pytest.org/en/latest/example/simple.html#pass-different-values-to-a-test-function-depending-on-command-line-options) of passing arguments into pytest functions that can be modified via the command line.


LiterallyUniqueLogin:
Sorry, I definitely should have included that explanation at the beginning. I thought tests with the code was nice too, but there's [this part](https://bioconda.github.io/contributor/guidelines.html#tests) of the bioconda docs explicitly forbidding shipping test data with the code, and this is the cleanest way I could manage to run tests without having the test data directly packaged with the code.

I also had the following short exchange on gitter:

LiterallyUniqueLogin
Hey all, I'm going to contribute to Bioconda for my first time, and I had some questions on testing

Specifically - the testing guidelines suggest not packaging test data with the code, which makes sense. But then how am I supposed to run meaningful tests at the end of a build to verify it has succeeded?

Devon Ryan
@LiterallyUniqueLogin We expect that tool authors run their own CI tests during development, so what you get after installation won't be any different. If you need additional testing then you can install the package locally before it's uploaded to ensure it's working correctly.

----

I think its important to run the tests in conda's build system - that's why there's an opportunity to run tests there. But I understand their reluctance to store large datasets with code - it slows down everyone's installations. I think this might be a happy medium?

Slimming down the test files would work too, but this honestly seemed easier and would allow us more flexibility in the future.

What do you think?

nmmsv:
Humm I think that makes sense. As long as we can have the tests in the repo and run them in travis (and potentially Conda), I think it's good. I do think it's good if users can also run the tests, which will give us more debugging power (as we don't have software testers, and user feedback can help us fix user-specific issues that we haven't seen).
Does it make sense to you?

LiterallyUniqueLogin:
Yeah, agreed.

Under this change, if you check out the git repo, you get all the test data too. Just running `pytest .` works.

If you install the source dist tarball vended on PyPI, you get the runtime code and the tests, but not the test data. Running `test_trtools.sh` will download the test data from the repo into a temporary location and then run pytest on it.

If you install the built distribution (the wheel on PyPI) which is the default for either conda or pip, then you'll get the minimal installation - runtime code but no tests or test data. Users who installed this way would need to pull down the git repo to get the tests.

Do you like this setup? Can I push the code?

nmmsv:
I think that should be fine, we can def play around with it after you
pushed (try different scenarios) and change it if needed.
Thank you!
---------------------
Nima Mousavi
PhD Candidate, GymrekLab <https://gymreklab.github.io/>
Electrical and Computer Engineering
UC San Diego
www.nimamousavi.com <http://nimamousavi.com>

nmmsv:
did you mean to close or merge? :D

LiterallyUniqueLogin:
Yeah, I opened a new PR.

PR #78, opened 2020-05-28
Improve io and errors

nmmsv:
Bigger changes:
1. Fixing auto detection of vcf type in compareSTR
2. Created a general purpose vcf reader and vcf type inference function in utils and used them in mergeutils as well.


nmmsv:
I was developing in OSX, but no idea why travis breaks in Linux tests. I'll fix it tomorrow.

PR #79, opened 2020-05-28
updated equal sign

richyanicky:
changed the = to ==


LiterallyUniqueLogin:
[Here](https://pip.pypa.io/en/latest/reference/pip_install/#example-requirements-file) is the packaging authority's example requirements file. This looks correct. Ship it!

PR #80, opened 2020-05-28
Change project structure to allow tests to be vended to consumers

LiterallyUniqueLogin:
This update does a few major things to the project structure:
* It moves the different command line tools (e.g. `mergeSTR`), which used to be their own top-level python packages, into subpackages of `trtools`. This was not necessary, but makes the build process cleaner and simpler to maintain, and it also allows for removing all the python path muckery in each file.
* It moves all of the tests into the python packages instead of being in their own folder in the repo. This is line with both how major python packages [numpy](https://github.com/numpy/numpy) and [dask](https://github.com/dask/dask) structure their projects. This allows our consumers to get our tests, and run them if desirable. This is particularly important for bioconda which will rebuild our package in its environment, and should run our tests during that rebuild. This change is necessary if we want to distribute the tests because the python packaging authority has specifically [removed](https://setuptools.readthedocs.io/en/latest/setuptools.html#non-package-data-files) the ability to distribute arbitrary files that reside outside of python packages, as their final locations in the user's environment would be different for different installs.
* It excludes the `sample_regions` and `sample_vcfs` directories from our distributions so that they are not too large and don't encumber our consumers. (This is required by the bioconda [guidelines](https://bioconda.github.io/contributor/guidelines.html#tests), and the rationale makes sense). In their place, this adds a `test_trtools.sh` script that which allows users who have installed trtools to download those files and then test trtools. I have set up bioconda to use this script during its build process. Note that users who download the source repo from git won't be affected by this - they will still get all the test files.
* Note: to run the tests on the git source repo, you'll now need to run `python -m pytest` from the root of the repo instead of just `pytest`. The latter will no longer work as it does not add the current directory to the python path and so no code will have access to the `trtools` package.

Some small changes include making sure the License file is included in both the source and built distributions we vend, and changing the tests to use a temporary folder provided by pytest instead of using the common/dumps folder that sticks around forever.

I am getting a small build-system error on my bioconda build with this set of changes (it can't find git, which is used by `test_trtools.sh` to download the test files). I don't know how to debug that, and I don't think it's necessary to fix that in this change, so I propose merging this in, and then asking the bioconda community about that when I open a PR with our build recipe with them.


nmmsv:
Oh almost forgot, can you merge this into develop? 😆

LiterallyUniqueLogin:
`test_trtools.sh` only works if trtools is installed. (You can install a local copy by doing `cd $REPO_ROOT ; pip install . ; test_trtools.sh`) Otherwise you should be running pytest directly (`cd $REPO_ROOT ; python -m pytest`) Do those work for you?

PR #85, opened 2020-06-17
Doc cleanup

LiterallyUniqueLogin:
### Changes, from most to least important
* The contributor documentation is cleaner than before and closer to our current best practices. I wrote this before Melissa released [this draft](https://docs.google.com/document/d/10jLBapZmGsg1ZNShYgzWq4E7eTAz3xINpmlx-_Zt0jw/edit), so there are still some discrepancies, but those can be fixed later, I think this set of edits (if we agree with them) are still cleaner than what was in the repo previously.
  * Note: My personal opinion is that our policies should be documented in text/rst files in the repository. If we're writing up changes to our policies for discussion (as opposed to just talking things through on a call), I think that those changes should be submitted as PRs to the repo, and we can discuss them on the PR thread. 
* No duplicate `TRTools` at the top of the readthedocs main page, but keep it there for the github.
* Formal python type hints now look exactly like docstring typehints in the docs: the difference in the docs for see tr_harmonizer.HarmonizeRecord as an example - the `trtools.utils.tr_harmonizer.Types` moved down into the Paramaters section of the function docs and leaves the function declaration more legible.
* Rename the VCFTYPES enum to VcfTypes and make it a proper enum class - this makes its own docs look cleaner, and makes it cleaner as a parameter type in function calls throughout the docs.
* Sphinx no longer throws any warnings during the build process
* Set version and release sphinx options correctly (though I don't know what these do ... still, seems better than having them specifically incorrect)


LiterallyUniqueLogin:
@nmmsv You need to have all the dependencies of trtools installed in order to build the docs. You can install them one by one, or `pip install -r requirements.txt` from the root of the repo (though then you need to make sure that the versions in that file don't conflict with the versions of anything in your environment - conda can help with that if that's an issue).

More easily - readthedocs now builds the documentation for each pull request. On this page there is a line saying "Read the Docs build succeeded!" - click on the details button next to that to see the docs for this PR.

nmmsv:
oh my bad, I forgot I'm running these on my mac and I don't have a bunch of these packages there.
It makes sense now, and I think the changes look good! thanks!

PR #86, opened 2020-06-19
Add quality score plots to qcSTR

LiterallyUniqueLogin:
Please review [my doc change PR](https://github.com/gymreklab/TRTools/pull/85) first, this PR is diff'ed against that change, assuming its already been accepted.

Now can graph quality scores. Please read qcSTR docs to figure out the specifics of the feature I've added - if those don't fully clear it up, then I should improve the docs. I shouldn't be explaining that here.

Changes to tr_harmonizer:
* TRRecord now asks for a quality score field in its construct
* it also asks for an optional transform that transforms that field into a 0-1 probability. Otherwise it assumes it is already a 0-1 probability.
* In particular, the quality scores popstr uses are weirdly truncated, so I have to do an awkward transform of them. See the note under GetQualityScore.
* ExpansionHunter doesn't provide quality scores, gangSTR has an option to disable them. hipSTR, popSTR and adVNTR always seem to provide them.
* class TRRecordHarmonizer was moved lower in the tr_harmonizer.py file, but it's contents were not changed aside from adding the quality related methods.

Other changes:
* Fixed bugs in qcSTR - crashing awkwardly when the output location is a directory, or when it's located in a misspelled directory. Don't crash on sample filtering typos. Specify that sample filtering typos are ignored.
* Encourage qcSTR users to use `bcftools view` first if they have complicated filtering desires.
* print updates to stdout as qcSTR produces plots instead of making the entire run long and silent
* add dependency on scikit-learn - this may be useful for some graphs, and its nice to add the dependency now instead of later once people are already consuming the package.
* There's no reason to recommend specifying `vcf-type` even if we allow it.

Question:
Can we remove the sample filtering feature, and just tell people to use `bcftools view`? That will save us some maintenance headache, and the sample feature has some weird corner cases already that aren't well documented.


nmmsv:
Thank you for making this tool, it looks like it has a ton of features and there are a lot of tests around the code to support it!
I think the changes look good to merge. Regarding your question, if the issues of the feature are too much we can do as you suggest. One argument against it is that in larger datasets users rather run the QC tool with different parameters than generate a lot of subsampled VCF files (which may be impossible in consortium size projects). let me know what you think.

Another comments: should we add a sentence to the docs to mention that QC plots may not be directly comparable across tools? Especially since different tools have different math models and different expected QC curves.

gymreklab:
I am merging this. But I agree with @nmmsv that:

- It would make sense to mention that these plots may not be comparable across tools.
- For sample filtering (I assume this refers to the --samples option?) I would like to keep that for exactly the reason pointed out: sometimes we have large VCF files with many samples but only want to analyze a subset. e.g. if we have families, we might want to only analyze parents. I wouldn't want to have to make a copy of a very big file and would prefer to specify a sample list.

LiterallyUniqueLogin:
Now merged into develop. Will handle the comments soon.

PR #88, opened 2020-06-24
Redux: Updates based on reviews

LiterallyUniqueLogin:
A redo of [this PR](https://github.com/gymreklab/TRTools/pull/84), now readthedocs will autobuild the doc changes so we can see them.

---

Updates to documentation, and some of the IO (Error handling)


LiterallyUniqueLogin:
I've made the following changes to Nima's branch: (most to least important)

* Fixed bugs and wrote tests for when the program used to crash ungracefully when:
  * Some, but not all of simultaneous records had contigs not present in the header
  * Records aside from the first one in a VCF had contigs not present in the header
* Change error messages regarding wrong contigs to correctly discuss the error at hand.
* Provide links and point users to TR caller documentation if they have additional questions about any specific caller.
* Grammatical fixes and docs cleanup
* Removed some clauses in the code that inappropriately would catch all errors instead of just the exception they were expecting.

nmmsv:
The changes look good to me! thank you for reopening the PR and adding the new changes!

PR #91, opened 2020-06-30
changing qcSTR threshold for refbias plot

gymreklab:
Minor change to require more data points for plotting the refbias plot in qcSTR.


LiterallyUniqueLogin:
This doesn't seem right. Why do we even have a threshold here at all? If I want to see what the average difference is for a variant with 24 or 99 samples, why shouldn't I be able to? I definitely wouldn't expect it to be filtered without anything in the docs mentioning that. One thing we could do is instead of binning in groups of 5 and using a piecewise linear function, we could do some sort of [weighted spline](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.UnivariateSpline.html) with weights proportional to the number of samples at each reference length.

Also: the plot produced by this says 'median', but as far as I can tell the mean is being calculated. We should fix that. I would prefer doing something like a different smoothing spline for each median, 90th quartile and 10th quartile and leave out the mean. If we plot the mean, I still think we should plot the median. If not, we need to mention in the docs that this plot is not robust and should only be run after filtering (e.g. qcSTR viewing, then dumpSTR).

gymreklab:
Thanks for these comments:
* You're right we say median but it is mean. I'm fixing the label. I see your points with mean/medians but for most datasets I've tried the median is pretty boring (0 in all bins).
* The threshold of 100 is pretty arbitrary. But we do need a way to exclude or alleviate the issue that bins with small number of data points will not behave nicely. I propose making options to set the min number of data points, the x-axis range, and whether to use mean or median. These can be set to reasonable defaults, and will honestly probably be rarely (although I find it useful to look at biases in different callsets which is why I want to keep it).
* I agree this isn't super robust, since the appropriate numbers to use are dataset specific. I can add to the docs saying that.

LiterallyUniqueLogin:
I just made a bunch of changes, both in reaction to the changes you made, and in reaction to the conversation today. Can you look over them? My commit messages should have the laundry list of what I've changed. You can see the diff of what I've changed against your last revision [here](https://github.com/gymreklab/TRTools/compare/df682b282c41f5b2f7c84bb30f6e996c847c5836...qcstr-smallfix)

gymreklab:
Thanks for these changes. All look good to me. If it looks good on your end you can go ahead and approve/merge. Thanks!

PR #92, opened 2020-06-30
clarifying mergeSTR docs

gymreklab:
Clarifying misleading mergeSTR summary at the top of the page, which previously suggested mergeSTR should be used to merge VCFs from multiple TR genotypers. In fact, mergeSTR requires the VCFs being merged to be created by the same genotyper.

### DIFF
--- a/trtools/mergeSTR/README.rst
+++ b/trtools/mergeSTR/README.rst
@@ -1,5 +1,5 @@
 .. overview_directive
-.. |mergeSTR overview| replace:: MergeSTR merges VCF files from differing TR genotyping tools into a single VCF file.
+.. |mergeSTR overview| replace:: MergeSTR merges multiple VCF files containing TR genotypes into a single VCF file.
 .. overview_directive_done
 
 MergeSTR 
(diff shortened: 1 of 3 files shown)


LiterallyUniqueLogin:
What does "if the allele definitions are different across different files" mean? Should that be "if different alternate alleles are present in the different files"? (If so, I'll confirm this by writing a test for that).

Or is this a holdover from claiming that mergeSTR can merge files from different genotypers, and should just be deleted?

gymreklab:
This was a holdover. I updated to your suggestion. If all looks good please approve and merge. Thanks!

LiterallyUniqueLogin:
Melissa, I made a few changes on top of your doc changes that are related, or just part of me trying to understand if what we say in our documentation is actually true. If you like the changes, feel free to merge.

A few changes:
* MergeSTR seems in really bad shape. I have lots of complaints about the design and the testing. I've now listed those at the bottom of the mergeSTR tests file.
* Make clear in two more places that mergeSTR does not merge VCFs from different genotypers
* Don't list which genotypers we support here - just one more place we have to update
* We don't have per-genotyper options, don't pretend to
* Better links
* Better --vcfs argument syntax
* Actually support the eh SVTYPE INFO field
* Clean up the GetSampleInfo function so I can understand it

gymreklab:
Thanks for the changes. I have looked them over and will merge.

Thanks for adding the TODOs to the tests. We should look at and discuss those later in a separate PR. I agree with many but not all.

FYI One major reason we specify which format fields we support is that for some (in particular, genotype likelihoods, but any field that has a number of things specified by the number of alternate alleles), are not trivially mergeable. So for now we wrote out the ones we know should make sense even after merging to make sure we don't fill in FORMAT fields with garbage. It's possible we could instead merge all fields of a certain type (e.g. those with a constant number of entries) and not merge others (e.g. per-allele or per-genotype fields like GL. But that is still dangerous. GangSTR's GGL field meets neither of those because it is a special case. So we can't just blindly append.

I'm not sure where we ever promised per-genotyper options. We do mention supporting different fields per genotyper which is accurate.

PR #93, opened 2020-06-30
added message to reheader if contig info missing for mergeSTR

gymreklab:
MergeSTR was previously updated to provide an error message when chromosomes were not present in the contigs specified in the VCF header. I have clarified this message to:
* Say which chromosome is missing, and print out the list of contigs we did find in the header
* Point to the bcftools reheader command which can be used to solve that issue.


LiterallyUniqueLogin:
I added the name of the file with the offending record to the error message you wrote, and then fixed a bug and wrote a few more tests around other issues related to contigs.

If you're happy with this, go ahead and merge.

gymreklab:
This looks fine to me. Thanks for the changes.

PR #95, opened 2020-07-02
Adding example files and commands with Platinum Genomes trio for all supported tools

gymreklab:
This PR adds example data files, and test commands, for files from all 5 supported genotypers. This primarily affects:

* `example-files` has new files for the Platinum Genomes trio for chr21 for each of the callers (adVNTR currently is an exception. It has dummy samples for now but will be updated and retested when that’s available)
* the `README.rst` files have been edited to include example commands at the bottom.

In the process of testing all of these, a variety of mostly minor changes were also made.

MergeSTR:
* When we check VCF types for files being merged, we were exploding we were confused that a file could come from more than one possible VCF type, even if the user had specified `—vcftyp`. This came up when testing a HipSTR file using a GangSTR reference. Both “HipSTR” and “GangSTR” were in the header of the VCF so we were confused. But even when I specified `—vcftype hipstr` it failed because `InferVCFType` failed. I have added `vcftype` as an optional argument to `InferVCFType`. If we detect more than one type, but the user-specified type is in that set, we return the user-specified type.
* Minor error message change to make it clear we’re quitting in mergeSTR if we fail to write the header (e.g. due to duplicate sample names across files).
* When merging things with conflicting refs, I changed to skip those records and output a warning rather than explode. This could be desirable if most of the records, but not all, are mergeable.

StatSTR:
* Was assuming “END” field present in VCF files. Which is not the case for PopSTR. Instead, compute END as POS+len(ref_alele)
* Output “.” for allele freqs/counts if there are no samples.

qcSTR:
No changes. But see other PR fixing issue with error "The directory  which contains the output location does not exist"

dumpSTR:
* Fixed similar issue with inferring VCF type when confused between multiple types.

compareSTR:
No code changes made. Note the advntr/popstr examples are not interesting as they are hard to compare (both to themselves, since I haven’t figured out how to run them yet to generate VCFs with different params for the same samples, but also to other files for similar reasons, like that I’d need to use a custom reference to make calls comparable). But at least the trivial examples of comparing a file against itself works for those.


gymreklab:
Edit: I have updated the AdVNTR examples to be based on the Platinum Genomes trio.

PR #96, opened 2020-07-04
Updating info on supported genotypers

gymreklab:
This is updated to match Supp Table 1.
Note, there is a link to contributing on the bottom that doesn't seem to get parsed correctly. I'm not sure about this syntax so hope one of you can look at that.


LiterallyUniqueLogin:
Some questions:
You say "designed to be run genome-wide". Is this not true of any of these? What does it mean for this not to be true?
PopSTR seems like it doesn't even use a reference except for super long repeats (based on my cursory reading of their docs). Why doesn't it get a * also?
What does "genic hg19" mean?

gymreklab:
I added back a bunch of (now edited) text to the use cases column. I do think this is important to have, since the main point of having this page is giving people that don't know a sense of what these tools were designed to do.

I do like having the info from before in separate columns. Thanks for adding. I don't know if you have a trick for dealing with RST tables. But it is not one of my favorite activities...

Not all of these are designed to run genome-wide. (although all could be). AdVNTR is currently too slow to reasonable run genome-wide (I forgot the run time, but it's in the hundreds of hours I think even for only things inside of genes ("genic").
ExpansionHunter was designed for targeted genotyping of expansions. If you have a complex locus with many interruptions, there is currently no tool to automate inferring the correct regular expression to use. SO in that sense that part is not (yet) automated to run genome-wide.
PopSTR has "markerInfo" files with the info for each genome-wide TR. But they don't give instructions for how to make a different one if you want to. (e.g. like, I don't see a specification of the format of that file).

But to give all the tools the benefit of the doubt, I've changed the asterisk thing and instead just note at the bottom that all could theroetically be run on any panel.

AdVNTR usually infers length. It can be used to infer frameshifts mutations. So that's not quite inferring sequence so I don't think it should get that label.
For EH, while complex TR sequences can be specified, as far as I can tell it only outputs lengths, which is why I put "length" there.

All of this info is hard to boil down in table form. But I think the columns you added as well as the use cases column gives some idea of intended uses of the different tools.

If it looks ok with you feel free to merge. Once we're happy and merged I'll update the supp text.

(Also sorry some of the commit messages are just "Update CALLERS.rst". Generally I like more informative messages. When editing on the web browser I clicked too fast through the commit button).

LiterallyUniqueLogin:
Hopefully this is the last round of changes. Just some minor stuff. If you like it, merge it.

Split the table in two for visual clarity.
Only differentiate between STRs and VNTRs where necessary, otherwise just change to TRs.
Say tools apply to all TRs despite the name.

PR #97, opened 2020-07-07
Adding Platinum Genomes Examples

gymreklab:
This PR includes:

* Adding two new Platinum Genomes Vignettes (making use of HipSTR, EH, and PopSTR)
* Small fixes how mergeSTR handles adVNTR and EH fields since those broke downstream commands after merging
* Cleaning up READMEs for each tool. Including:
  * Remove REPODIR etc. syntax and assume people are testing from a pip install
  * Alphabetize example commands based on the tool used
  * Cleaned up some links
  * Cleaned up syntax of describing inputs/outputs to be more consistent across the readmes for the different tools.


LiterallyUniqueLogin:
I see when adding example files in the previous PR you added cryptically named c57 and ceu files to /example-files but don't use them anywhere. Can you either rename them so their use/origin is obvious or delete them.

More review to come

gymreklab:
Those files were there before. At least some get used in the vignettes so I prefer not to remove.

gymreklab:
Will need to update with new QC CDF with 0 to 1 x-axis. Wait to merge this until that change from #101 is merged

LiterallyUniqueLogin:
Comments based on add-pg-examples branch

Can we have a README in example-files/ saying where all the files were downloaded from, or if they were generated, what commands we used to generated them. Doesn't need to be before submission, but soon after so we don't forget it?

In dumpSTR:
I don't understand why the column AdVNTRCallMinDepth is being emitted with value zero test_dumpSTR_advntr.samplog.tab
It's weird that in the examples mean samples per passing locus < 1, that means we're keeping loci with zero calls. Maybe have a file with more than one sample, and then we can filter all loci with no calls?
It seems that some of the filters we're running in these examples aren't filtering any TRs. Can we use more realistic filters instead? The null case isn't a very good example case.

More comments to come later.

gymreklab:
Added comments clarifying dumpSTR output. Readme for example files can be done later. I'm not sure it's necessary. These are meant purely as examples to test TRTools functionality, not to give a full tutorial on each of the genotypers supported.

LiterallyUniqueLogin:
Merging before review so we can release and resubmit for publication. Will continue to review after release.

PR #101, opened 2020-07-08
Qc str fixes

LiterallyUniqueLogin:
Done:
* quality plots swapped back to 0-1 and update plots accordingly
* Reminders on qcSTR on how to use quality plots
* Explanation of quality plot defaults in qcSTR
* Page on how TRHarmonizer interprets quality scores from different genotypers
* Moved API page from sidebar of the readthedocs to a subpage of the library page
* Fix a bug where PHRED score 0 calls could have lower quality than PHRED score 1 calls
  in popSTR VCFs
* Fix a bug where qcSTR progress updates to the command line were caught in a buffer
  and not printed out in time.

To do later today about quality and qcSTR:
* Note in compareSTR to use qcSTR for quality plots (and to read the note in qcSTR).
* Link from qcSTR to the quality page in the Library docs.


LiterallyUniqueLogin:
Oh whoops. I changed it to produce pngs for the docs, and forgot to change it back to PDFs which is our default. I'll switch it when I'm back.

gymreklab:
I changed png->pdf, but there is still an error being raised at `assertFEquals(trh._ConvertPLtoQualityProb([0, 1, 1, 1]), 0)`

LiterallyUniqueLogin:
Yep, I'll fix that test when I get back.

gymreklab:
I don't think we're interpreting the PL field correctly. PL gives phred-scaled likelihoods of each genotype. There is no reason the likelihoods should add to 1 (these are likelihoods, not posteriors). So the function that does 1-sum of non-zero things wouldn't be valid.

Options are:
- Alternatively we could covert to a Q score comparable to GangSTR/HipSTR by returning the likelihood of the best genotype divided by the sum of likelihoods of all possible genotypes. This wouldn't be too far off of what we have. Would have to update _ConvertPLtoQualityProb and docs on that
- don't support quality for popSTR for now

LiterallyUniqueLogin:
I believe the PL field is actually a probability function despite having 'likelihood' in its name. If so this should be correct. If it's a true likelihood function, I don't think there is any relationship between the likelihoods at different model parameters and so I don't think we can normalize. I think we'd have to abandon this trick and just bin all PHRED score 0 likelihoods at probability == 1. 

Not positive about my interpretation here. I sent out an email to the vcf-spec community and cc'ed you on it. They were helpful to me before, so I'm hoping they can clarify.

To move forward, I'll remove probability scores for popSTR for now.

LiterallyUniqueLogin:
New commit should fix all outstanding issues related to qcSTR or quality scores that need to be done before resubmission.

gymreklab:
Edited some minor wording. ALso commented unused phred functions for now. Merging

PR #102, opened 2020-07-08
fixed HWE error for haploid

ileenamitra:
X chrom / haploid cannot handle HWE calculation


gymreklab:
Merging so we don't through an error on this for now. But we should submit an issue to fix HWE (along with other issues arising with chrX/Y).

PR #108, opened 2020-07-11
Dockerfix

LiterallyUniqueLogin:
I've checked, the docker file builds. To confirm: we both don't have testing around the dockerfile, nor do we distribute it, right? So there's no need to merge to master and publish?


LiterallyUniqueLogin:
Whoops, made my commit at the wrong point in the timeline.

LiterallyUniqueLogin:
Nvm, develop was just lagging behind. The commit is still good, so please review.

PR #109, opened 2020-07-11
Fix docker installation issues do to install trtools improperly

LiterallyUniqueLogin:
I've checked, the docker file builds. There's no testing around this, right? And we don't distribute the dockerfile, right? So there's no rush to merge this to master.


gymreklab:
I prefer this: #111 since we may want to modify to pull other branches which wont be available on conda or pip

PR #110, opened 2020-07-11
StatSTR fixes and entropy statistic

LiterallyUniqueLogin:
Implement entropy statistic. I want a statistic that measures locus complexity like heterozygosity for some looks at the UKB data, but I believe this is a better one. Please read what I say about it in the docs, see if you are okay with the wording.

General cleaning up of statSTR docs. Be explicit about what happens in nocall situations (it was confusing seeing example files in them where the only row's output was '.' - it made me question whether I understood the output format correctly) Unfortunately github's diff shows me rewriting everything, but most lines its just a few word changes. Space examples out for better readthedocs viewing.

I also moved function types from comments to python type hints in a few places - they're equally clear to the readthedocs, but better for IDE tooling that I'm using (it lets me look up function definitions automatically, for instance)


gymreklab:
I would remove all of the justification of the entropy metric in the docs (the stuff in parentheses). It's arguable if it's a better metric than heterozygosity, which has a meaning more directly tied to popgen theory. But in any case, in the description of the options is not the place to argue which is better, but simply to say what it is just like we do for other metrics.

I won't change the syntax of the function definitions. But it was confusing to me here and elsewhere since I hadn't seen that used before and since we use it on some functions but not others throughout TRTools. We also advertise that we use numpy docstrings, which I like and want to keep, and these no longer conform to that.

Otherwise looks good thanks for adding.

LiterallyUniqueLogin:
I've removed the comparison between entropy and heterozygosity, but still kept some explanation because I do not expect most users to understand this option without an explanation.

Type hints are a [standardized](https://www.python.org/dev/peps/pep-0484/), if optional, python function syntax. And specifying types is optional in numpy docs, so we still conform. I'd like do this in the entire package, but that takes time I don't want to put in right now, so I've been doing it bit by bit.

gymreklab:
Ok, if Guido's on board then I guess I'll take it. merging

PR #111, opened 2020-07-11
upgrading docker to ubuntu 20.04

gymreklab:
Upgrading Docker to use Ubuntu 20.04, which has python 3.8


LiterallyUniqueLogin:
Assuming you've checked that this builds, it looks good.

PR #116, opened 2020-07-29
Trying to fix set is not ordered errors from Travis runs

LiterallyUniqueLogin:
Errors are found in [this](https://travis-ci.org/github/gymreklab/TRTools/builds/713077113) travis build which was caused by merging a few bugfixes from master into develop (not sure how, maybe updating the dockerfile? Does travis even use that file?)


LiterallyUniqueLogin:
I wasn't able to reproduce this locally (here's some specific version of Pandas that caused the issue and I couldn't figure out the version) so I didn't add any new tests around this, but it's now passing.

PR #118, opened 2020-08-01
Update utils.py and statSTR to use cyvcf2

LiterallyUniqueLogin:
Please only review this once [the tr_harmonizer PR](https://github.com/gymreklab/TRTools/pull/117) has been accepted and merged, otherwise the diff will include those changes too.

All statSTR tests and utils tests now pass.

Other changes:
* statSTR now has tests that confirms it produces the same output as a previous run that's been saved (output taken from the PyVCF version of statSTR). Before none of the tests actually looked at the output, they just confirmed it did/did not crash.
* statSTR now prints out diagnostics to the terminal telling you how fast it's running per sample.
* Sped up statSTR testing.
* Made precision configurable (and reduced default value) and made mode deterministic, to allow for the reproducibility test.


trtools/utils/utils.py
```diff
@@ -6,73 +6,70 @@
...
-        List of VCF files to merge
-    checkgz: bool, optional
+    vcf_locs :
+        A list of vcf locations
+    checkgz:
         Check if each VCF file is gzipped and indexed
-    region : str, optional
-        Chrom:start-end to restrict to
-    
+        If any are not, return None
```
nmmsv, line 62: Does this mean that if a file is not gzipped, checkgz is going to be None?
I think the name and the functionality of the variable don't match really well. But I'm guessing this is not a recent change, so we can fix it after this round of updates.

nmmsv (approved):
It all looks good to me! Thank you for making the changes!

trtools/utils/utils.py
LiterallyUniqueLogin: It means that if checkgz is True (the default), then if any of the VCFs are not gzipped, none of them are returned. (None is returned instead). Otherwise, a list of cyvcf2.VCF objects is returned.

If it is False, then a list of VCFs will be returned if all of them can be found, regardless of if some were or were not gzipped.

Happy to clean this up at the end.

PR #121, opened 2020-08-06
Because of a change a while back, the tools can no longer be run as s…

LiterallyUniqueLogin:
…tandalone scripts, they must be installed locally. Update the docs and tools to reflect that.

I moved Contact Us to above Contributing because Contributing is getting really big and doesn't apply to most people, so people reading from top to bottom might stop reading there.

I didn't change any wording in the  `#. Document your changes.` section, I just indented things by an extra space as I've realized is proper and deleted an extraneous empty line. But the PR diff doesn't seem to understand that.


LiterallyUniqueLogin:
The code coverage is failing here because I deleted a few lines of code, not because I added any untested code.

nmmsv (approved):
I tested locally and the changes look good to me! thanks!

PR #127, opened 2020-12-19
Port mergeSTR to cyvcf2

LiterallyUniqueLogin:
Please only do this review once statSTR, qcSTR and dumpSTR have been reviewed and merged into the cyvcf2 branch.

Functional changes are documented in RELEASE_NOTES.rst , other than that the code should work as before, just on cyvcf2 and faster. This seems like one of the modules that will benefit the least from the speed up - I still expect this code to be incredibly slow and not usable at large scales. Happy to discuss thoughts on speeding it up if that's of interest.

There are now tests for AdVNTR and Expansion Hunter and tests that check the output of the code. There is now a utils file in testsupport where the VCF comparison code has been moved to so tests from multiple tools can use it.

As part of this change, I removed some of the rude comments I'd written previously in one of the MergeSTR files. Dredging that up wasn't the point of this review, but I didn't want to leave that there. Apologies for that.


trtools/dumpSTR/tests/test_dumpSTR.py
```diff
@@ -6,6 +6,7 @@
 import vcf
 
 from ..dumpSTR import *
+from trtools.testsupport.utils import assert_same_vcf, assert_same_file
```
nmmsv, line 11: Is this file meant to be deleted? It's in dumpSTR

nmmsv (approved):
I went over them and made a few minor comments. I tried it on my system and it works as expected. Thanks for making these changes!

trtools/mergeSTR/mergeSTR.py
```diff
@@ -129,47 +97,61 @@ def get_sources(reader):
...
     # Write sample list
-    samples = mergeutils.GetSamples(readers, usefilenames=args.update_sample_from_file)
+    try:
+        if not args.update_sample_from_file:
+            samples = mergeutils.GetSamples(readers)
+        else:
+            filenames = [fname.split('/')[-1] for fname in args.vcfs.split(',')]
+            samples = mergeutils.GetSamples(readers, filenames)
+    except ValueError as ve:
+        common.WARNING("Error: " + str(ve))
```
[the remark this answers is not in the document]
LiterallyUniqueLogin: I disagree. The entire program exits if this happens, so I think that's a good candidate for it being an error.

trtools/dumpSTR/tests/test_dumpSTR.py
LiterallyUniqueLogin: The entire file isn't deleted, just those methods, which have been moved to trtools/testsupport/utils.py

PR #128, opened 2020-12-29
Cyvcf2 comparestr

LiterallyUniqueLogin:
This should be the last cyvcf2 PR done.

The tests for compareSTR are quite sparse - no tests currently check its output. They should probably be written at some point. But there's still a few more tests than previously

There were lots of user facing changes with this update, mainly because CompareSTR was very buggy. Those changes are detailed in RELEASE_NOTES.rst. Please go over them to make sure those choices seem reasonable.

Thanks!


nmmsv (approved):
Hi Jonathan, I think the changes that were made and the assumptions for IO look good to me. I wasn't able to go into the details of code changes in the method. I think since this is an auxiliary method (and not a thoroughly tested one), we need future developments to focus on the main method to add features as we need them, and add tests as we encounter bugs.
One thing that was missed is the version bump. Currently, it's at 3.0.2.

PR #151, opened 2022-02-15
Compare str comparison improvements

Kulivox:
Changes in this pull request are based on suggestions from this comment 
https://github.com/gymreklab/TRTools/pull/147#discussion_r803773789
My changes change what determines whether two records are comparable. Before, everything was based on function, which provided a list that said which records should be skipped in the next iteration. If this list was only made up of True values, records were compared. I've decided to change that, this list is still produced, but another, more configurable function determines the comparability of the two records.


trtools/compareSTR/tests/test_compareSTR.py
```diff
@@ -173,6 +174,90 @@ def test_main(tmpdir, vcfdir):
...
+    assert not handler(records, chrom_idxs, min_idx)
+
+    records = [DummyHarmonizedRecord("chr1", 10, 4, "AC"), DummyHarmonizedRecord("chr1", 10, 4, "AC")]
+    chrom_idxs = [np.inf, 0]
+    assert not handler(records, chrom_idxs, min_idx)
+
+    chrom_idxs = [0, 0]
+    assert handler(records, chrom_idxs, min_idx)
+
+    records = [DummyHarmonizedRecord("chr1", 10, 5, "AC", "rec1"), DummyHarmonizedRecord("chr1", 10, 4, "AC", "rec2")]
```
LiterallyUniqueLogin, line 250: For completeness sake, let's also add a test where the end coordinates are the same but the start coordinates differ

trtools/compareSTR/compareSTR.py
```diff
@@ -713,12 +736,86 @@ def bad_range():
...
                            "start-end where end <= start")
             return 1
-    except ValueError: # an int() cast failed
+    except ValueError:  # an int() cast failed
         return bad_range()
 
     return 0
 
 
+def handle_overlaps(args: Any = None) \
```
LiterallyUniqueLogin, line 745: Until we have a defined implementation for any value of args aside from None, I'd rather not have this here, and just have the handler function. Because if we never implement any other uses for handle_overlaps then this just leaves the code in a more confusing state. And if we do implement another use case here, it should be easy to add handle_overlaps() back in.

LiterallyUniqueLogin (commented):
Looking good! Just two comments

trtools/compareSTR/compareSTR.py
Kulivox: I agree. I've removed this function.

trtools/compareSTR/tests/test_compareSTR.py
Kulivox: I've added the test

trtools/utils/mergeutils.py
```diff
@@ -248,13 +248,20 @@ def GetMinRecords(record_list: List[Optional[trh.TRRecord]], chroms: List[str])
     return [CheckPos(r, chroms[min_chrom], min_pos) for r in record_list]
 
 
+def GetCoordinatesOfRecord(record: Optional[trh.TRRecord]) -> Tuple[int, int]:
+    if record is None:
+        return np.inf, np.inf
 
-def GetMinHarmonizedRecords(record_list: List[Optional[trh.TRRecord]], chroms: List[str]) -> List[bool]:
-    r"""Check if each record is next up in sort order
+    return record.pos, record.pos + record.ref_allele_length * len(record.motif)
```
LiterallyUniqueLogin, line 255: I'm thinking that we should add a field called 'end_pos' to the TRRecord that has this number in it `record.pos + record.ref_allele_length * len(record.motif)` and then just access that here. Do you agree that field may be useful in more than just this context?

trtools/compareSTR/tests/test_compareSTR.py
```diff
@@ -173,6 +174,90 @@ def test_main(tmpdir, vcfdir):
...
+    args.vcftype2 = 'hipstr'
+    retcode = main(args)
+    assert retcode == 0
+    with open(tmpdir + "/test_compare-locuscompare.tab", "r") as out_overall:
+        lines = out_overall.readlines()
+        # Two of the records wont be compared
+        assert len(lines) == 2
+        _, err = capfd.readouterr()
+                        ## first output is about records that have the same starting position but different end pos
+        assert err == ("Records STR_40 and STR_40 overlap:\n"
```
LiterallyUniqueLogin, line 213: For both of these records, the start of the records are the same (if you properly ignore the flanking bps). Can we have one where the starts are different but the ends are the same? It should still say that they overlap but are not comparable.

trtools/compareSTR/tests/test_compareSTR.py
```diff
@@ -173,6 +174,90 @@ def test_main(tmpdir, vcfdir):
...
+    records = [DummyHarmonizedRecord("chr1", 10), None]
+    min_idx = 0
+
+    assert not handler(records, chrom_idxs, min_idx)
+
+    records = [None, DummyHarmonizedRecord("chr1", 10, 4, "AC")]
+    chrom_idxs = [np.inf, 0]
+    assert not handler(records, chrom_idxs, min_idx)
+
+    records = [DummyHarmonizedRecord("chr1", 10, 4, "AC"), DummyHarmonizedRecord("chr1", 10, 4, "AC")]
```
LiterallyUniqueLogin, line 242: This is a situation that should never come up, right, where there's a non-None record with chrom index np.inf? If that's the case, I think I'd like to remove this test - it's just confusing for people trying to read it later and infer what use case this is testing.

LiterallyUniqueLogin (commented):
Noticed a few things that I should've caught in the first go round, but hopefully they're not a big deal to address.

trtools/utils/mergeutils.py
Kulivox: Yes, I agree. I've also noticed that this: `record.pos + record.ref_allele_length * len(record.motif)` is off by one, so I added end position attribute to TRRecord and fixed this problem

trtools/compareSTR/tests/test_compareSTR.py
Kulivox: You are right, np.inf is not possible in this situation. I've updated the test so it tests the same functionality but simulates a possible case.

trtools/utils/tr_harmonizer.py
```diff
@@ -680,6 +682,15 @@ def __init__(self,
             self.has_fabricated_ref_allele = False
             self.ref_allele_length = len(ref_allele) / len(motif)
 
+        # declaration of end_pos variables
+        self.end_pos = int(self.pos + self.ref_allele_length * len(motif) - 1)
+        self.full_alleles_end_pos = self.end_pos
```
LiterallyUniqueLogin, line 687: A few comments:
* I would rather not define this here and then redefine it later if we have full_alleles, that could lead to confusion for a reader. Let's just have "if full_alleles is None then .... else ...."
* If we're going to define this, can we document this (in the same way you documented end_pos). It can go in the attributes section. 
* It seems `full_alleles_pos` is already defined but not documented, do you mind documenting that likewise?

trtools/utils/tr_harmonizer.py
```diff
@@ -680,6 +682,15 @@ def __init__(self,
             self.has_fabricated_ref_allele = False
             self.ref_allele_length = len(ref_allele) / len(motif)
 
+        # declaration of end_pos variables
+        self.end_pos = int(self.pos + self.ref_allele_length * len(motif) - 1)
```
LiterallyUniqueLogin, line 686: This cast is unsafe, `int(1.9999)` goes to 1, and it's possible that with finite precision this doesn't properly evaluate to an exact integer.  Let's use round instead. We can have a comment here explaining why we're using round when all we want is a cast because in theory it should be a whole integer.

trtools/compareSTR/compareSTR.py
```diff
@@ -713,12 +736,64 @@ def bad_range():
...
+    if chrom_indices[0] != chrom_indices[1] or \
+            chrom_indices[0] != min_chrom_index or \
+            chrom_indices[1] != min_chrom_index:
+        return False
+
+    left_start, left_end = left.pos, left.end_pos
+    right_start, right_end = right.pos, right.end_pos
+
+    overlap = min(left_end, right_end) - max(left_start, right_start)
+    # This calculation contains max() - 1 to compensate
```
LiterallyUniqueLogin, line 782: I don't think this is the right implementation. Shouldn't we be adding one to overlap above instead of subtracting one from the denominator?

trtools/utils/mergeutils.py
LiterallyUniqueLogin: Thanks

LiterallyUniqueLogin:
Changes all seem to be going in the right direction!

trtools/compareSTR/compareSTR.py
Kulivox: Yes, you are right, this implementation only calculates the right score for completely overlapping sequences, otherwise, it would calculate a score that is slightly lower than intended. Thank you for pointing that out!

trtools/utils/tr_harmonizer.py
Kulivox: I've applied your suggestions.

LiterallyUniqueLogin:
Sorry for letting this hang - life got busy. This looks done! Please add an 'unreleased changes' section to RELEASE_NOTES.rst in the main directory with a bullet or two about what's changed here and I'll merge this into develop.

PR #153, opened 2022-02-15
Merge str harmonization update

Kulivox:
Similar update to the one done here https://github.com/gymreklab/TRTools/pull/147
MergeSTR is now able to correctly handle HipSTR records which may or may not have their flanking bps as part of the record


trtools/mergeSTR/mergeSTR.py
```diff
@@ -172,19 +179,44 @@ def GetRefAllele(current_records, mergelist):
     ref : str
        Reference allele string. Set to None if conflicting references are found.
     """
-    refs = []
+
+    def DefaultKey(record: trh.TRRecord):
```
LiterallyUniqueLogin, line 183: I believe harmonization should make it so that record.ref_allele is correct regardless of the vcftype, so you don't have to do this sort of check here. So you should just be able to pull that field. (Or if not, we should fix the harmonization to make that so).

trtools/mergeSTR/mergeSTR.py
```diff
@@ -172,19 +179,44 @@ def GetRefAllele(current_records, mergelist):
...
     chrom = ""
     pos = -1
     for i in range(len(mergelist)):
         if mergelist[i]:
-            chrom = current_records[i].CHROM
-            pos = current_records[i].POS
-            refs.append(current_records[i].REF.upper())
+            chrom = current_records[i].chrom
+            pos = current_records[i].pos
+            refs.append(ref_picker(current_records[i]).upper())
```
LiterallyUniqueLogin, line 200: When this comes from the tr harmonizer, it'll already be uppercase

LiterallyUniqueLogin (commented):
We need tests running mergeSTR on actual HipSTR VCFs to confirm that it properly handles records with flanking base pairs. The current tests don't confirm that there's been any upgrade in functionality. I think you can build tests taking appropriate records from the `example-files/*hipstr*` files, but if that doesn't work I'm happy to provide example HipSTR output.

trtools/mergeSTR/mergeSTR.py
Kulivox: You are right, when I wrote this I was afraid that the REF field might contain something like <STR41>, similarly to ALT fields from some callers, but now I realize that something like that wouldn't make any sense. Thanks for pointing that out!

Kulivox: I reverted the changes but found an additional problem. It looks like EH records usually have only one BP as a value in the REF column. Harmonization of EH records expands the EH record motif and fabricates this REF allele. Use of ref from the harmonized record can cause differences like this: 
```
Output line: CAGCAGCAGCAGCAGCAGCAGCAGCAGCAGCAGCAGCAGCAGCAGCAGCAGCAGCAG
Control line: C
```
A possible solution is to use the old REF field from the non-harmonized VCF record and to use the old REF and ALT fields from non harmonized records (reverting most of my changes), but this might cause some weird behavior, for example on HipSTR records with and without flanking BPs.
But I guess it depends on what you think is the intended behavior. Is it OK to remove the flanking BPs during the merge or not?

LiterallyUniqueLogin: Yes, we should remove flanking bps during the merge and make the explicit in the documentation.

I had forgotten about the different representations between callers issue. You're right, it doesn't make sense use harmonized records for outputting in general, it needs to look like the original caller's output. I think the solution in the current commit is fine then - check if it's a HipSTR record, and if so use the harmonized record's alleles which have the flanking bps removed (and are sequences which is the same type of output that HipSTR produces). Otherwise just proceed with what's in the record, as this problem only pertains to HipSTR.

That all seem reasonable to you?

trtools/mergeSTR/tests/test_mergeSTR.py
```diff
@@ -34,6 +34,16 @@ def __init__(self, chrom, pos, ref, alts=[], info = {}):
         self.ALTS = alts
         self.INFO = info
 
+
+class DummyHarmonizedRecord:
```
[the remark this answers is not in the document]
Kulivox: Yes, this is just a temporary solution.

trtools/mergeSTR/mergeSTR.py
Kulivox: Conversion to upper case now only happens when non harmonized REF is used

trtools/mergeSTR/mergeSTR.py
Kulivox: Yes, it is. I've reverted my local changes and fixed the case problem from your other comment.

trtools/mergeSTR/tests/test_mergeSTR.py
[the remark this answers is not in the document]
Kulivox: I've also created a new test that confirms that flanking bps of hipstr records are ignored, and the repeat part is merged as before.

LiterallyUniqueLogin:
This looks done! Please add an 'unreleased changes' section to RELEASE_NOTES.rst in the main directory with a bullet or two about what's changed here and I'll merge this into develop.

PR #159, opened 2022-06-30
fix the minimum position bug in mergeutils

yli091230:
The GetIncrementAndComparability function in mergeutils.py gives error message "stuck in infinite loop and exiting error". The error is caused by the min_pos may get from STRs on different chromosomes when samples have different called STRs. Therefore generate an increment list with all FALSE values. Fix it by get the min_pos from the chromosome== min_chrom_index only


LiterallyUniqueLogin:
Hi Yang,

I ran the pytests and trtools/mergeSTR/tests/test_mergeSTR.py::test_GangSTRRightFile is failing. Can you fix that? (Or confirm that it's an issue with how I'm running it). Also, you put a random 'G'  into the .gitignore file which I think wasn't intended.

yli091230:
> 

Hi Jonathan,

Thanks for reviewing the code. 
1. What error message you got from the  trtools/mergeSTR/tests/test_mergeSTR.py::test_GangSTRRightFile? 2. 
2. I think I might accidentally pressed on keyboard when I editing the .gitignore file. I'll fix that.
Thanks,
Yang

PR #164, opened 2022-08-12
Add support for Beagle imputed VCFs

LiterallyUniqueLogin:


yli091230:
Hi Jonathan, I got the following error after running pytest. 
```
trtools/dumpSTR/tests/test_dumpSTR.py::test_zippedOutput - FileNotFoun...
trtools/dumpSTR/tests/test_dumpSTR.py::test_TwoDumpSTRRounds - FileNot...
trtools/dumpSTR/tests/test_dumpSTR.py::test_BrokenVCF - TypeError: Rec...
```
The first two "FileNotFound" are due to the missing tabix tool under my environment, I'm not sure do we need to add a note that tabix should be installed, which might be a common sense. 
The third one is trtools/utils/tr_harmonizer.py:294: TypeError. Not sure whether it is a issue with the vcd file or the way how I'm running it. 

I'll keep go through the details of the changes.

LiterallyUniqueLogin:
Hey Yang, I've addressed your comments!

yli091230:
Thanks. I think it looks great! I add a line in the `trtools_prep_beagle_vcf.sh` script to remove `$temp_file` if the scripts failed, feel free to remove it if that not make sense. 
Two questions about the documentation:
1. It seems the there are a few parameters in TRtools are not supported if data is from beagle, do we need to point out those in the notes?
2. I can see in the reference file you provided contains `INFO` fields like `START END PERIOD`. I could find those fields/annotation from the reference files downloaded from our [lab website](https://gymreklab.com/2018/03/05/snpstr_imputation.html). Do we need provide how to add those annotation to the reference file?
Thanks,
Yang

trtools/dumpSTR/dumpSTR.py
```diff
@@ -419,6 +441,9 @@ def CheckFilters(format_fields: Set[str],
         if vcftype != trh.VcfTypes["eh"]:
             common.WARNING("ExpansionHunter options can only be applied to ExpansionHunter VCFs")
             return False
+        elif is_beagle:
+            common.WARNING("ExpansionHunger call level filters cannot be applied to Beagle VCFs")
```
gymreklab, line 445: ExpansionHunger->ExpansionHunter (I always accidentally make this typo!)

trtools/utils/tr_harmonizer.py
```diff
@@ -1582,6 +1611,17 @@ def HasQualityScore(self) -> bool:
...
 
+    def IsBeagleVCF(self) -> bool:
+        """
+        Is this a VCF produced by running the Beagle software to impute STRs from a panel generated by an TR genotyper?
+
+        See Also
+        --------
+        tr_harmonizer.IsBeagleVCF
+        """
+        global IsBeagleVCF
```
gymreklab, line 1622: globals make me nervous. do we need to use a global variable here?

gymreklab (commented):
Thanks for going through this effort! I have added some minor comments to specific lines.

trtools/utils/tr_harmonizer.py
LiterallyUniqueLogin: I believe I was incorrectly using this. I thought this was necessary to access the file-scoped function defined outside the class. But this doesn't seem to be the case, so I'm just removing this statement.

LiterallyUniqueLogin:
> Thanks. I think it looks great! I add a line in the `trtools_prep_beagle_vcf.sh` script to remove `$temp_file` if the scripts failed, feel free to remove it if that not make sense. Two questions about the documentation:

Thanks for pointing this out! You're right that the script should act cleanly when its dependencies aren't present.

It's important to figure out what the script should do when it meets an error condition, which is not normally to continue on silently. Here the script needs tabix in order to run properly, so it should error out when tabix is not present. I've changed the code to reflect that.

> 1. It seems the there are a few parameters in TRtools are not supported if data is from beagle, do we need to point out those in the notes?

There's a brief line in the dumpSTR doc saying that we don't have any call-level filters for Beagle at the moment. Do you have any other parameters/notes in mind?

> 2. I can see in the reference file you provided contains `INFO` fields like `START END PERIOD`. I could find those fields/annotation from the reference files downloaded from our [lab website](https://gymreklab.com/2018/03/05/snpstr_imputation.html). Do we need provide how to add those annotation to the reference file?

Yep! Will email generate the files needed to fix this and email you separately.

>    Thanks,
>    Yang

I appreciate the review!

@mgymrek I believe I've addressed all your comments

PR #183, opened 2023-07-14
fix: vcf input via stdin

aryarm:
allows `trtools` to accept piped VCFs or VCFs that are specified via process substitution

Previously, the user would simply receive this error message, instead:
> Could not find VCF file

### DIFF
--- a/trtools/utils/tests/test_utils.py
+++ b/trtools/utils/tests/test_utils.py
@@ -1,4 +1,4 @@
-import io, os, sys
+import os, sys
 import numpy as np
 import pytest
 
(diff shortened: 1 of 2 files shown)


aryarm:
@aarushi03 does this help with the issue you were having?

LiterallyUniqueLogin:
This seems fine. I don't know if it actually does what it intends to (there are no tests regarding standard in, and I haven't dug enough to check for myself), but the new test is good and I can't see this breaking anything a user is likely to run into.

aryarm:
Thanks for the review!

I tried to create a test with stdin and wasn't able to. I just don't know how to do that in a way that allows `cyvcf2` to read from it

My only external validation that we have that this works is that @aarushi03 manually checked it

aryarm:
Thanks for the review!

I tried to create a test with stdin and wasn't able to. I just don't know how to do that in a way that allows `cyvcf2` to read from it

My only external validation that we have that this works is that @aarushi03 manually checked it

PR #184, opened 2023-07-16
refactor!: remove pybedtools as a dependency. This slightly changes the interpretation of BED files. From now on, the end position of each region will not be included.

aryarm:
TRTools previously used pybedtools to read tabix-indexed BED files, but it turns out that pysam can do that already

since this was the only use for pybedtools in the codebase, this PR also removes it from the list of dependencies

this PR doesn't add any tests because the relevant code is already tested in [`test_RegionFilter()`](https://github.com/gymreklab/TRTools/blob/e0ed1ec62808b3a19596c0d60a30127cf6f0473f/trtools/dumpSTR/tests/test_filters.py#L197-L232)

Once this PR gets merged, it might be easier to package TRTools with haptools somehow. I was previously unable to do it because (I think) pybedtools is one of those packages on PyPI that hasn't published it's dependency metadata and thus is impossible to install with `poetry`? (see [here](https://github.com/statsmodels/statsmodels/issues/8200) for another example and more info)
```
> curl -s https://pypi.org/pypi/pybedtools/json | jq ".info.requires_dist"
null
```

@gymreklab, as the original author of the region handling code, can I request your review? Feel free to let me know if there are more tests I should include


aryarm:
ok, one of the dumpSTR tests did in fact fail 😦  I'm not sure why...

It appears as if the failing tests are those where files output by dumpSTR are compared to those that have been already generated. There are two lines that are slightly different. For now, I just changed the already generated files to match what they should be. See 9784fecfeb7f53bbf2afe6de3c0c99a67417c1d2 for the changes. Hopefully that's ok, @gymreklab ?

LiterallyUniqueLogin:
That's fine. IIRC, those tests were a quick way to prevent future bugs from going unnoticed without exhaustively testing the module which is more desirable but very time intensive. If the change in output is acceptable, just swap the old file for the new. The test could also be removed if it's causing more issues than helping, but hopefully not.

trtools/testsupport/sample_vcfs/dumpSTR_vcfs/locus_filters.vcf
```diff
@@ -439,7 +439,7 @@ chr21	14383027	GANGSTR_437028	TTTTGTTGTTGTTGTTG	TGTTGTTGTTGTTGTTG	.	HETLOW0.05;f
...
+chr21	14405799	GANGSTR_437032	ACCCACCCACCC	ACCCGCCCACCC	.	HWE0.5;HETHIGH0.45;foo_region	INFRAME_PGEOM=0.95;INFRAME_UP=0;INFRAME_DOWN=0;OUTFRAME_PGEOM=0.95;OUTFRAME_UP=0;OUTFRAME_DOWN=0;START=14405799;END=14405810;PERIOD=4;NSKIP=0;NFILT=0;BPDIFFS=0;DP=342;DSNP=0;DSTUTTER=0;DFLANKINDEL=1;AN=6;REFAC=3;AC=3;HRUN=3;HET=0.5;HWEP=0.25	GT:GB:Q:PQ:DP:DSNP:DSTUTTER:DFLANKINDEL:PDP:PSNP:GLDIFF:AB:DAB:FS:ALLREADS:MALLREADS:FILTER	0|1:0|0:1:0.5:115:0:0:1:59.46|55.54:0|0:179.46:-0.11:104:-0.16:0|87:0|86:PASS	0|1:0|0:1:0.5:109:0:0:0:53.35|55.65:0|0:152.82:-0.04:98:-0.53:0|83:0|81:PASS	0|1:0|0:1:0.5:118:0:0:0:74.97|43.03:0|0:123.77:-2.66:104:0:0|82:0|78:PASS
```
LiterallyUniqueLogin, line 442: The PR seems good overall. Can you just explain briefly why the output is changing? Was it wrong in the first place?

aryarm: ah, oops - see [my (now updated) original comment](https://github.com/gymrek-lab/TRTools/pull/184#issuecomment-1803002481) about it. I think pysam might treat end coordinates as exclusive of the region while pybedtools treats them inclusive. But it could also be due to other things like 0-based vs 1-based indexing

I'm going to convert this PR back into a draft until I have some time to look through this and figure out what's going on

LiterallyUniqueLogin: Okay, I'm okay with that. Officially, bed files are end-point exclusive, so I don't mind treating them that way, that seems right, we should just specify that in our docs.

LiterallyUniqueLogin:
This change looks good, I don't see anything else you need to review (though feel free to if you'd like). If you want to push, I'll approve. Just mention in our README that bed files are end point exclusive, and mention this in the change log.

LiterallyUniqueLogin:
We need to retitle this PR to make it clear to the version-handling software that this is a breaking change. Will approve after that!

aryarm:
done! I added an exclamation to the conventional commits prefix

LiterallyUniqueLogin (approved):

PR #186, opened 2023-08-01
Updated the test files for mosaicSTR and simTR

PatoPena:
Wrote tests to the mosaicSTR and simTR codes to verify they work correctly.


gymreklab:
Thanks for these test cases!
Because of the rename (from mosaicSTR to prancSTR) I was unable to merge automatically and had to manually copy these tests over to the new prancSTR files. So closing this PR.

PR #187, opened 2023-08-11
perf: speed up `TRRecord.GetLengthGenotypes()`

aryarm:
I was playing around with `GetLengthGenotypes` when trying to integrate it into haptools and noticed there might be a potentially faster way of implementing it.

This implementation vectorizes the for-loop using [broadcasting](https://numpy.org/doc/stable/user/basics.broadcasting.html), improving the speed as we scale up 1) the number of samples and 2) the number of alleles in each TR of a VCF. I benchmarked haptools for 1 (but not 2) via [our benchmarking script](https://github.com/CAST-genomics/haptools/blob/main/tests/bench_genotypes.py) (with some small adaptations).

<p align="center">
benchmark_GetLengthGenotypes
</p>

I didn't add any tests to this PR because it's just a refactor. `GetLengthGenotypes` already has tests and they all seemed to pass with this new code. But feel free to let me know if there are any tests that it would be good for me to add.


aryarm:
@LiterallyUniqueLogin, can I request a review from you since you were the original author of this code? What do you think of these changes?

trtools/utils/tr_harmonizer.py
```diff
@@ -1091,19 +1091,16 @@ def GetLengthGenotypes(self) -> Optional[np.ndarray]:
         idx_gts = self.GetGenotypeIndicies()
         if idx_gts is None:
             return None
+        else:
+            idx_gts = idx_gts.astype(int)
```
LiterallyUniqueLogin, line 1095: Why is this needed here? Were these floats, and using them as indices was failing before the cast? If so, should we move this line into the GetGenotypeIndicies function? Or do you believe it better to leave here?

LiterallyUniqueLogin (commented):
This looks good, thanks for the update! Just one minor comment to address.

trtools/utils/tr_harmonizer.py
aryarm: yes, exactly. They were `float64`s before. I'm not sure why...

I originally didn't make this change in `GetGenotypeIndicies` because I wasn't sure whether it would break anything. But the tests seem to pass after doing it in 1b43aa397e114f82b77eb11508f3c74f77254802, so I guess it's ok?

LiterallyUniqueLogin (approved):

PR #193, opened 2023-10-30
build: migrate from setup.py to pyproject.toml and set up CI

aryarm:
...since setup.py is going to be deprecated ([source](https://snarky.ca/what-the-heck-is-pyproject-toml/))

we also migrate from travis CI and code-cov to github actions according to [these directions](https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python). This will allow us to integrate testing into the PR approval process

Note: We still don't have a lock file or any version constraints in our pyproject.toml, so our CI is still prone to death whenever a breaking change to one of our dependencies is uploaded to PyPI. If we want to tackle that, it can be a future-me problem 😅 

note: this pr also fixes some issues arising from deprecations in the readthedocs config ([source](https://blog.readthedocs.com/use-build-os-config/)) and resolves all warning messages in the current readthedocs build. In order to get the tests to pass, I also had to make some small alterations to the code


aryarm:
@LiterallyUniqueLogin, can I ask you for approval for this, too?

it should also help to fix the failing build issues in #191

aryarm:
@LiterallyUniqueLogin, it looks like the tests are failing because `bgzip`, `bcftools`, and `tabix` aren't installed

> ERROR trtools/testsupport/sample_vcfs/associaTR/make_dosages.py - subprocess.CalledProcessError: Command 'bash -c "bgzip -f gp_dosages.tsv && bgzip -f ap1_dosages.tsv && bgzip -f ap2_dosages.tsv && bcftools annotate -a gp_dosages.tsv.gz -h gp_dosage_header.hdr -S <(tail -n +2 samples.txt) -c CHROM,FROM,TO,FMT/GP many_samples_biallelic.vcf.gz | bcftools annotate -a ap1_dosages.tsv.gz -h ap1_dosage_header.hdr -S <(tail -n +2 samples.txt) -c CHROM,FROM,TO,FMT/AP1 - | bcftools annotate -a ap2_dosages.tsv.gz -h ap2_dosage_header.hdr -S <(tail -n +2 samples.txt) -c CHROM,FROM,TO,FMT/AP2 - > many_samples_biallelic_dosages.vcf && bgzip -f many_samples_biallelic_dosages.vcf && tabix -f many_samples_biallelic_dosages.vcf.gz "' returned non-zero exit status 127.

Are those tools necessary for the associaTR tests? If so, I could try to install them in the CI environment.
If it's too difficult to install, is there a possibility that we could just skip whichever tests require those tools?

aryarm:
@LiterallyUniqueLogin, I was able to install the conda dependencies and the tests now pass! It's ready for you to review now!

trtools/__init__.py
```diff
@@ -1,2 +1,4 @@
-from .version import __version__
-
+try:
+    from .version import __version__
+except ModuleNotFoundError:
+    __version__ = "unknown"
```
LiterallyUniqueLogin, line 4: This is a bit concerning. When does this come up?

pyproject.toml
```diff
@@ -0,0 +1,55 @@
+[build-system]
+requires = ["setuptools", "setuptools-scm"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "trtools"
+authors = [
+    {name = "Melissa Gymrek", email = "mgymrek@ucsd.edu"},
+]
```
LiterallyUniqueLogin, line 9: If this is going to be here, we should add Nima Mousavi, myself, you and others(?) (Aarushi if her tool has been published) to the list.

pyproject.toml
```diff
@@ -0,0 +1,55 @@
...
+description = "Toolkit for genome-wide analysis of STRs"
+readme = "README.rst"
+requires-python = ">=3.5"
+license = {text = "MIT"}
+classifiers = [
+    "Development Status :: 4 - Beta",
```
LiterallyUniqueLogin, line 15: We can remove this.

pyproject.toml
```diff
@@ -0,0 +1,55 @@
...
+]
+dynamic = ["version"]
+
+[tool.setuptools]
+packages = ["trtools"]
+script-files = ["trtools/testsupport/test_trtools.sh", "scripts/trtools_prep_beagle_vcf.sh"]
+license-files = ["LICENSE.txt"]
+
+[tool.setuptools_scm]
+version_file = "trtools/version.py"
```
LiterallyUniqueLogin, line 41: It seems like this file has been deleted. So is this line wrong?

PUBLISHING.rst
```diff
@@ -42,12 +42,8 @@ Once changes have been made to develop that are ready to be published, first cho
 
 Then go through the steps of merging the changes into the master branch:
 
-#. Run :code:`pytest` and make sure all the tests pass. Then run :code:`./test/cmdline_tests.sh` and make sure those tests pass.
```
LiterallyUniqueLogin, line 45: I believe the user should still run the tests themselves locally and not just rely on the CI. I'm worried people won't know to think about testing and take ownership of that otherwise. So perhaps keep this line as is? Thoughts?

PUBLISHING.rst
```diff
@@ -42,12 +42,8 @@ Once changes have been made to develop that are ready to be published, first cho
...
 #. Change the 'Unreleased Changes' section of :code:`RELEASE_NOTES.rst` to the new version number.
 #. Check if any changes have been made that have not yet been documented in the release notes. If so, document them.
-#. Update the version number in setup.py
```
LiterallyUniqueLogin, line 48: If not the old way, how are we handling versioning now?

LiterallyUniqueLogin (commented):
Looking good overall! I haven't tested that any of the changes work as described, I've just read through the docs. Just a few minor comments.

Thank you for all the work to keep this package maintained!

PUBLISHING.rst
aryarm: sure, I'll add it back in

Why do you think the user should still run the tests themselves? imo it's helpful to be able to run the tests locally b/c you can debug things better that way, but I'm not sure it's really necessary to run it locally again right before you push it to github where it'll get run a second time

PUBLISHING.rst
aryarm: the way things are handled now is that when you run `pip install build`, setuptools will automatically create a `version.py` file at the path specified in the pyproject.toml
the version that it uses will be whichever one you used when you created the git tag earlier
I've added a clarification for this that explains the latter part

pyproject.toml
aryarm: actually, setuptools will use this path to figure out how to generate a `version.py` file for us at the moment when we build the tool

so this should be fine as it is

trtools/__init__.py
aryarm: it shouldn't. If at all, it _might_ happen during development (when you do `pip install -e`) - but even then, setuptools should generate a temporary version for us

aryarm: I can remove this line if you'd like, but I think it will make it easier for us to bypass any issues that might (but shouldn't) arise from setuptools not properly generating a temporary version number during development

pyproject.toml
aryarm: done

pyproject.toml
aryarm: ok, done!

trtools/__init__.py
aryarm: also, it *will* happen if someone does `pip install .` (without the `-e`) on a local clone of the repo, but nobody should be doing that, as far as I know

PUBLISHING.rst
LiterallyUniqueLogin: Because I can imagine other people in the lab not realizing that it is their job to make sure all the tests pass if they're not explicitly told. So they'll push to PR, forget about it, come back later and the tests will have failed, and then they'll go ask someone what happened, as opposed to being given immediate feedback that they wrote something which needs fixing.

For someone who knows what's going on, whatever workflow's good for them is fine. This line wouldn't be targeted at them.

LiterallyUniqueLogin (commented):
All looking good. One last change: I'd remove the mention of using the web interface to create tags under the section "git tagging" (and then also remove the mention that there are two ways). Reason being that it seems this process requires the tag to exist locally so that `python -m build` uses that tag, and the web interface wouldn't meet that requirement, right?

PUBLISHING.rst
LiterallyUniqueLogin: Thank you for explaining!

trtools/__init__.py
LiterallyUniqueLogin: Okay. Seems good.

aryarm:
Actually, if they use the web interface, the tag should still exist locally as long as they do a `git pull` afterwards, which is already one of the steps anyway. I'll add a message reiterating that the most recent commit needs to be tagged, just in case.

LiterallyUniqueLogin (approved):

PUBLISHING.rst
aryarm: that makes sense! sounds good

PR #199, opened 2023-12-16
build: remove Dockerfile, add codespace config, and bump min python

aryarm:
This PR removes our now-defunct Dockerfile. Instead, we recommend that you use the TRTools image maintained by the Bioconda team on the biocontainers registry. The docs have been updated to reflect this.

To offset the removal of the Dockerfile, this PR adds an alternative way to work with TRTools in the cloud: [GitHub codepsaces](https://docs.github.com/en/codespaces)! This can be helpful when you want to quickly play around with our code but don't want to go through all of the setup steps. Just click on the "Open in Github Codespaces" badge on the README and voila - you'll be presented with an environment automagically set up with TRTools and our development dependencies. This environment will run with (usually free) credits on Github's codespace VMs.

Finally, this PR also specifies a strategy for our support for various python versions. From now on, we will only formally support active versions of python that are not end-of-life (3.8+). However, at least for now, we think that TRTools will still run properly for versions 3.6 and 3.7 - although support for those versions is not guaranteed and may break at any time.


aryarm:
@aarushi03, can I request your review on this PR, since you've worked with GitHub codespaces before? Can you try out the new badge (checked out to the `build/Dockerfile` branch) and let me know if it works for you?

[](https://codespaces.new/gymrek-lab/TRTools?ref=build%2FDockerfile)

aarushi03:
> @aarushi03, can I request your review on this PR, since you've worked with GitHub codespaces before? Can you try out the new badge ([checked out to the `build/Dockerfile` branch](https://github.com/gymrek-lab/TRTools/tree/build/Dockerfile#trtools)) and let me know if it works for you?
> 
> [](https://codespaces.new/gymrek-lab/TRTools?ref=build%2FDockerfile)

Certainly. Reviewing it now.

.devcontainer/environment.yml
aarushi03, line 1: Will -conda-forge::python=3.8 only allow it to be compatible with python 3.8 or even higher versions of python?

aarushi03, line 1: Will -conda-forge::python=3.8 only allow it to be compatible with python 3.8 or even higher versions of python?

README.rst
```diff
@@ -78,6 +78,13 @@ Note: make sure TRTools is not installed in the environment via a different meth
 
 Note: if you will run or test :code:`simTR`, you will also need to install `ART <https://www.niehs.nih.gov/research/resources/software/biostatistics/art/index.cfm>`_. The simTR tests will only run if the executable :code:`art_illumina` is found on your :code:`PATH`. If it has been installed, :code:`which art_illumina` should return a path.
 
+With Docker
+^^^^^^^^^^^
+
+Please refer to `the biocontainers registry for TRTools <https://biocontainers.pro/tools/trtools>`_::
+
```
aarushi03, line 85: Add "Run the following command:"

aarushi03 (changes requested):

README.rst
```diff
@@ -78,6 +78,13 @@ Note: make sure TRTools is not installed in the environment via a different meth
 
 Note: if you will run or test :code:`simTR`, you will also need to install `ART <https://www.niehs.nih.gov/research/resources/software/biostatistics/art/index.cfm>`_. The simTR tests will only run if the executable :code:`art_illumina` is found on your :code:`PATH`. If it has been installed, :code:`which art_illumina` should return a path.
 
+With Docker
+^^^^^^^^^^^
+
+Please refer to `the biocontainers registry for TRTools <https://biocontainers.pro/tools/trtools>`_::
```
aryarm, line 84:
```suggestion
Please refer to `the biocontainers registry for TRTools <https://biocontainers.pro/tools/trtools>`_ for all of our images. To use the most recent release, run the following command::
```

.devcontainer/environment.yml
aryarm: yes! this ensures that python 3.8 is the version that gets used during development, which will reduce the chance of the developer introducing a change that breaks support for our lowest supported version

I've added a comment that explains that within the yml file

aarushi03 (approved):

PR #203, opened 2024-01-02
fix: add version constraints

aryarm:
I've opened this PR for us to develop minimum version constraints for our dependencies. I've made a first pass at this, which I hope to refine with the help of whoever reviews this! These constraints will reduce the chance that users encounter bugs when installing TRTools in environments that have old versions of our dependencies.

## my strategy

I started with the versions of each dependency that had just been published at the time of [the first commit to the repository](https://github.com/gymrek-lab/TRTools/commit/e2c8402b2f5a9e9abec017f1a22729515b92e0fe) (see 2b1f871885cc938c1a6183f9e8725b11331869ed). As expected, the tests did not pass when run with some of those versions. In general, support for old dependencies can be broken by ongoing code maintenance and new features.

So I incrementally bumped certain versions until the tests passed (see 4be0437aaf2c54232fd36ead3afe33df652d1074). Some combination of binary search through our dependencies' versions and ctrl+f through our dependencies' changelogs was helpful here. It was also super helpful to use poetry, a popular dependency management tool, to help me navigate version incompatibilities. Based on my experience, I think we will need to continue using it if we want to keep supporting specific python versions. (Refer to #202 for more details.)

## a bug requiring us to drop support for py 3.6

For example, there was one instance when I discovered that some of the tests (for code that was added to TRTools recently) would only pass for versions of pandas >= 1.2.0. I suspect that there must be some sort of bug in TRTools when used in combination with versions of pandas < 1.2.0. But when I tried to constrain pandas to >= 1.2.0, poetry informed me that I would be breaking support for python 3.6. Apparently, pandas dropped support for python 3.6 in 1.2.0.

So I had to drop TRTools' support for python 3.6, too. In any case, I don't think the version of setuptools that comes with python 3.6 supports pyproject.toml files ([source](https://pypi.org/project/ppsetuptools/)), so we would probably have to revert our pyproject.toml file to a setup.py file if we wanted to continue supporting python 3.6. Also, we will have to drop support for python 3.6 (and certain versions of numpy, cyvcf2, and pysam) anyway when we merge the `haptools.data` and `tr_harmonizer` libraries, since `haptools.data` only supports python 3.7+.


aryarm:
It's very likely that there are valid versions lower than the ones I began with. I could also have made a mistake when incrementing the version constraints. When reviewing this PR, you can run our tests against a lower version of a dependency via poetry and pytest. For example, to try numpy 1.17.2:

1. `poetry add 'numpy==1.17.2'`
2. `poetry run pytest`


I recommend doing this in [a Github codespace](https://github.com/features/codespaces) so that you don't mess up your local installation of TRTools. To create a codespace, click on "Create codespace" on [this page](https://github.com/codespaces/new/gymrek-lab/TRTools?skip_quickstart=true&machine=basicLinux32gb&repo=154568691&ref=deps%2Fconstraints&devcontainer_path=.devcontainer.json&geo=UsWest). Wait a bit for the codespace to get set up and then open a terminal ( Ctrl + \` ) and run `conda activate trtools` to access the environment installed with poetry.


After you're done, you can manually edit the pyproject.toml file to ensure each version of each dependency still has a `>=` sign in the front of it. And then you should also run `poetry lock --no-update` to ensure the lock file remains synced to the pyproject.toml file.

aryarm:
unless anyone has any objections, I'll probably merge this PR by EOD

Melissa and I agreed (via personal conversation) that these version constraints are probably low enough for now

LiterallyUniqueLogin (approved):

aryarm:
this pr also pins numpy to a version less than 2.0 for now, until we have a chance to test it and make sure it doesn't break anything

PR #201, opened 2023-12-19
ci: automate releases, add check for py3.7, and create pr checklist

aryarm:
This PR lets us automate the process of releasing new versions using [Google's Release Please Github action](https://www.padok.fr/en/blog/release-please). It works by creating (and maintaining) a _"release PR"_ which, when merged, will update the changelog, bump the version number, and upload our source distribution and wheel to PyPI.

## how is the version number determined?
The version number is determined automatically according to 1) [semantic versioning](https://semver.org/) and 2) the titles of each PR (which should follow [the conventional commits spec](https://www.conventionalcommits.org/)). For example, if all of the recently merged PRs are prefixed with `fix: `, then the system will infer that all recent changes have been bug fixes, so it will only bump the "patch" version. However, if even one of the recently merged PRs is prefixed with `feat: `, then the system will bump the "major" version because it will infer that a new feature has been added to `master`.
This makes our versioning more interpretable to our users while also redistributing the work of maintaining the version onto each developer of our toolkit.

## how is the changelog maintained?
The changelog for each release is simply the contents of the text body of the _release PR_. These text entries will themselves come from the titles of each of the PRs that have been merged since the last release. For example, [this](https://github.com/CAST-genomics/haptools/pull/198) is what the changelog entry would look like if the following two PRs were merged into master since the last release:
- fix: NoneType error in `Haplotypes.__iter__`
- fix: precision of phenotypes written to pheno file

[<p align="center"></p>](https://github.com/CAST-genomics/haptools/pull/198)

[Here](https://github.com/CAST-genomics/haptools/pull/223) is also a more recent example.

## some caveats
1. we will need to require that PRs get [squashed](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/about-pull-request-merges#squash-and-merge-your-commits) in order for this to work. imo this makes the git history [cleaner and easier to follow,](https://github.com/googleapis/release-please?tab=readme-ov-file#linear-git-commit-history-use-squash-merge) anyway
2. every PR that gets merged into master will have to have a prefix attached to it's title that follows [the conventional commits spec](https://www.conventionalcommits.org). For example, if the PR is a bug fix, it will need to be prefixed by `fix: `. Or, if it is a new feature, it will need to prefixed by `feat: `. I've added another Github action to prevent merging of PRs that don't have this
[...]

PR #204, opened 2024-01-03
Adding option to mergeSTR to give file with list of VCFs

gymreklab:
Currently mergeSTR requires the argument `--vcf` which takes a comma-separated list of VCF files to merge. When running mergeSTR on a large number of files, this results in very long commands. This PR adds the option `--vcfs-list` to mergeSTR as an alternative which allows users to specify a file with a list of the VCFs to merge (one file per line).

### DIFF
--- a/trtools/mergeSTR/tests/test_mergeSTR.py
+++ b/trtools/mergeSTR/tests/test_mergeSTR.py
@@ -14,6 +14,7 @@ from trtools.testsupport.utils import assert_same_vcf
 def args(tmpdir):
     args = argparse.ArgumentParser()
     args.vcfs = None
+    args.vcfs_list = None
     args.out = str(tmpdir / "test")
     args.update_sample_from_file = False 
     args.quiet = False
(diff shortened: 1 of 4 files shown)


trtools/mergeSTR/tests/test_mergeSTR.py
```diff
@@ -46,6 +47,20 @@ def __init__(self, chrom, pos, ref, alts=None, info=None):
...
+    fname2 = os.path.join(mrgvcfdir, "test_file_gangstr2.vcf.gz")
+    args.vcftype = "gangstr"
+    listfile = str(tmpdir / "test.list")
+    f = open(listfile, "w")
+    f.write(fname1+"\n")
+    f.write(fname2+"\n")
+    f.close()
+    args.vcfs_list = listfile
+    args.vcfs = None
+    assert main(args)==0
```
aryarm, line 62: in addition to checking that the exit code is non-zero, it might also be good to check that the output appears like you would expect via `assert_same_vcf()`?

```suggestion
    fname1 = os.path.join(mrgvcfdir, "test_file_hipstr1.vcf.gz")
    fname2 = os.path.join(mrgvcfdir, "test_file_hipstr2.vcf.gz")
    args.vcftype = "hipstr"
    listfile = str(tmpdir / "test.list")
    f = open(listfile, "w")
    f.write(fname1+"\n")
    f.write(fname2+"\n")
    f.close()
    args.vcfs_list = listfile
    args.vcfs = None
    assert main(args) == 0
    assert_same_vcf(args.out + '.vcf', mrgvcfdir + "/hipstr_merged.vcf")
```

aryarm (commented):
looks good! I made a small suggestion for the test

also, we should remember to list this change in the `CHANGELOG.md` file once we merge #201

aryarm (approved):

trtools/mergeSTR/tests/test_mergeSTR.py
gymreklab: thanks, this is a great suggestion! I added a check like this to the test

PR #208, opened 2024-01-23
fix: deprecation of binom_test in scipy 1.12.0 and cryptic error occurring with incorrect VCFs in cyvcf2 0.30.26

### LINKED ISSUE #206: broken test `test_dumpSTR.py::test_BrokenVCF`
## the problem
With the newest cyvcf2, `test_dumpSTR.py::test_BrokenVCF()` fails:

```
Exception: error parsing variant with `htslib::bcf_read` error-code: 0 and ret: -2
```
## bug hypothesis
This test is actually supposed to check whether dumpSTR will properly fail when given a malformed VCF. The VCF used in the test has a variant record with only four fields instead of the required ten.

So I think what's happening is that htslib is catching the error and reporting it to cyvcf2 before TRTools has a chance to catch it and exit gracefully with a non-zero exit code. The exception is happening when we call `next(self.vcffile)` here:

https://github.com/gymrek-lab/TRTools/blob/1d61fd1011dd3f6e0cf1c41cb21a32b8b8be208c/trtools/utils/tr_harmonizer.py#L1622

## potential solutions
Our options:

1. Catch the exception within `tr_harmonizer.py` and somehow gracefully handle it. This will probably require changing all of the code that iterates over VCFs using `TRRecordHarmonizer` to catch and gracefully handle exceptions arising from within it.
2. Change the test so that it succeeds regardless of whether an exception is raised. In this case, we could catch the exception within `tr_harmonizer.py` and then reraise the exception with a more useful message, since TRTools users might be confronted by that exception from now on.
3. Ask htslib or cyvcf2 to handle this upstream. I think this exception is probably a desired behavior for them, though.

My vote is for option 2

aryarm:
This PR handles errors that arise when you upgrade to the newest versions of our dependencies, specifically for cyvcf2 and scipy

It resolves #206 and also handles the deprecation of `scipy.stats.binom_test` in favor of `scipy.stats.binomtest`: https://docs.scipy.org/doc/scipy/release/1.12.0-notes.html#expired-deprecations

Merging this PR will resolve failures in the CI checks for all of our other PRs

### DIFF
--- a/trtools/dumpSTR/dumpSTR.py
+++ b/trtools/dumpSTR/dumpSTR.py
@@ -1201,6 +1201,13 @@ def main(args):
                 return 1
             else:
                 raise te
+        except ValueError as ve:
+            message = ve.args[0]
+            if 'properly formatted' in message:
+                common.WARNING("Could not parse VCF.\n" + message)
+                return 1
+            else:
+                raise ve
         if args.verbose:
             common.MSG("Processing %s:%s"%(record.chrom, record.pos))
         record_counter += 1
--- a/trtools/qcSTR/qcSTR.py
+++ b/trtools/qcSTR/qcSTR.py
@@ -501,7 +501,24 @@ def main(args):
 
     # read the vcf
     numrecords = 0
-    for trrecord in harmonizer:
+    while True:
+        try:
+            trrecord = next(harmonizer)
+        except StopIteration: break
+        except TypeError as te:
+            message = te.args[0]
+            if 'missing' in message and 'mandatory' in message:
+                common.WARNING("Could not parse VCF.\n" + message)
+                return 1
+            else:
+                raise te
+        except ValueError as ve:
+            message = ve.args[0]
+            if 'properly formatted' in message:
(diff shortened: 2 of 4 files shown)


aryarm:
@LiterallyUniqueLogin, I decided to catch the error in all downstream tools (which turned out to just be dumpSTR and qcSTR), since it was easy enough to implement

Let me know if there are any other changes we need!

aryarm:
@LiterallyUniqueLogin, I'm realizing now that this error will probably crop up in any tool that uses cyvcf2, and not just in dumpSTR and qcSTR.

The proper way to handle this is probably to add a BrokenVCF-equivalent test to every tool, and then use those tests to figure out how to properly catch the exceptions. But tbh those sorts of changes would probably bring this PR beyond what I have the bandwidth for. Also, I don't think I'm familiar enough with TRTools to be able to write those tests.

LiterallyUniqueLogin:
Your reasoning sounds good. Fixing this shouldn't be held up by missing tests for other utilities.

LiterallyUniqueLogin (approved):

PR #214, opened 2024-02-19
fix: clone command in `test_trtools.sh` testing script

aryarm:
This fixes a mistake I made in b23fa5635f7a10f871e57b6d08640f7a9025fcbc (in PR #202). I forgot to clone into the _current_ directory and only realized when the bioconda test failed in https://github.com/bioconda/bioconda-recipes/pull/45894#issuecomment-1953220461. We managed to patch the bioconda recipe, but anyone who tries to run the `test_trtools.sh` script after installing TRTools 6.0.0 with pip will still run into an error. This PR fixes that error for the next release of TRTools.

**Update (2/20):** I also added commit 37c75bb to ensure the `test_trtools.sh` script can be executed locally in a development installation of TRTools. Of course, this should only be used to test that the `test_trtools.sh` script itself is working. It won't actually test TRTools properly because it will use an outdated version of our test and data files.

### DIFF
--- a/trtools/testsupport/test_trtools.sh
+++ b/trtools/testsupport/test_trtools.sh
@@ -2,13 +2,13 @@
 
 # Run the tests for an installed copy of trtools
 # If not already present, megabytes of test data will be downloaded before test running
+# Please note that this script tests TRTools against
+# test files committed to the repo and will miss any local changes to those files
 
 # If TRTools was installed normally, this script can be executed by simply running
 # 'test_trtools.sh' on the command line.
 # Otherwise, if TRTools was installed via poetry, you should run the command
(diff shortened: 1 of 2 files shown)


trtools/testsupport/test_trtools.sh
```diff
@@ -2,7 +2,13 @@
 
 # Run the tests for an installed copy of trtools
 # If not already present, megabytes of test data will be downloaded before test running
+# Please note that this script tests TRTools against
+# test files committed to the repo and will miss any local changes to those files
```
aryarm, line 6:
```suggestion
# test files from the last published release and will miss any local changes to those files
```

PR #218, opened 2024-06-05
ci: fix broken GitHub actions

aryarm:
Our pull request checks broke because of https://github.com/readthedocs/readthedocs.org/issues/11150 and https://github.com/actions/runner-images/issues/9741 ! Let's fix them


aryarm:
merging without review

PR #220, opened 2024-06-18
fix: support for numpy 2.0

aryarm:
Since the resolution of  https://github.com/brentp/cyvcf2/issues/307, all of our dependencies now have support for numpy 2.0! Let's remove the upper bound requirement

This PR also makes some small changes to the release workflow to allow us to use [trusted pypi publishing](https://docs.pypi.org/trusted-publishers/). Because of these changes, we will need to set up trusted publishing before we publish a new version of trtools. Otherwise, the release workflow will break upon an attempted publication


LiterallyUniqueLogin (approved):
I haven't read enough about trusted publishing to understand what that is, but this all seems reasonable. You have my go ahead!

aryarm:
thanks!! Trusted publishing is just a more secure way for us to publish to pypi

Also, the new system uses github environments, so that admins of trtools will need to approve each pypi release

PR #227, opened 2024-08-22
feat: support for Apple M1 silicon

aryarm:
cyvcf2 and pgenlib added support for m1 apple silicon. See https://github.com/bioconda/bioconda-recipes/pull/50219 and https://github.com/bioconda/bioconda-recipes/pull/50220. So can TRTools be run on macOS ARM now? Are there any other dependencies that fail with the newest macs?

Let's add a test to our CI for this. If it works, then we can add osx-arm64 to the "additional platforms" that are supported in our bioconda recipe


aryarm:
Tests are failing because bcftools and art don't have macOS ARM support

But bcftools got it in 1.20 (see https://github.com/bioconda/bioconda-recipes/pull/49104), so we just need to figure out art

aryarm:
@nicholema , can you try to install TRTools on your mac again? 

```
conda env create -n trtools-conda -c conda-forge -c bioconda 'bioconda::trtools'

nicholema:
Thank you for adding this feature! It worked as,
conda env create -n trtools-conda -c conda-forge -c bioconda
'bioconda::trtools'

nicholema (approved):

PR #228, opened 2024-08-24
docs: generate internal links dynamically

aryarm:
readthedocs hosts different versions of our documentation. The bottom/left corner of their site allows you to select and change between the versions:

<p align="center"></p>

The `stable` version of our docs refers to the latest version that has been released, while the `latest` version refers to whatever is currently on the `master` branch.

Usually, if a user is browsing the `latest` version of the docs, all of the internal links will also point to the `latest` version. And vice versa for the `stable` version. But unfortunately, many of the pages of our docs have direct links to an internal URL rather than dynamically generated links.

This PR fixes that so that all internal links are generated dynamically. The drawback of this is that the links will no longer work when the docs are previewd on GitHub directly, but I think we usually prefer to preview the docs on readthedocs anyway.


gymreklab (approved):
thanks for doing this! this does move all the docs out of the main readme which is what people see on the github landing page, but points them to the appropriate place in readthedocs so should be fine

PR #232, opened 2024-08-30
feat: Add longtr support

gymreklab:
This PR adds explicit support for LongTR VCF files, meaning most commands will now accept `--vcftype longtr`. Because [LongTR](https://github.com/gymrek-lab/longtr) is an extension of HipSTR for long reads, the format is very similar to HipSTR and some HipSTR functions are reused. The main difference is the stutter fields are not present. 

In fact, in most cases giving a LongTR VCF but specifying `--vcftype hipstr` should still work fine. 

## Checklist

* [x ] I've checked to ensure there aren't already other open [pull requests](../../../pulls) for the same update/change
* [x ] At the top of the PR, I've [listed any open issues that this PR will resolve](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword). For example, "resolves #0" if this PR resolves issue #0
- [x ] I've explained my changes in a manner that will make it possible for both users and maintainers of TRTools to understand them
* [x ] All directories with large test files are listed in [the "exclude" section](https://python-poetry.org/docs/pyproject/#include-and-exclude) of our pyproject.toml so that they do not appear in our PyPI distribution. All new files are also smaller than 0.5 MB.
* [ x] I've updated the relevant REAMDEs with any new usage information and checked that the newly built documentation is formatted properly
* [ x] (if applicable) I've updated the pyproject.toml file with any changes I've made to TRTools's dependencies, and I've run `poetry lock --no-update` to ensure the lock file stays up to date and that our dependencies are locked to their minimum versions


doc/CALLERS.rst
```diff
@@ -58,6 +61,8 @@ For more information on a genotyper, please see its website linked below.
 +----------------------------+--------------------------+----------------------------+------------------------+--------------------------+-------------------------+------------------------+
 |    PopSTR_ (v2.0)          | 1-6bp                    | Yes                        | Length                 | 540,1401 (hg38)          | Illumina                | Many                   |
 +----------------------------+--------------------------+----------------------------+------------------------+--------------------------+-------------------------+------------------------+
+|    LongTR_ (v1.0)          | 1+bp                     | No                         | Length, sequence       | No ref provided          | PacBio, ONT             | Many                   |
```
heliziii, line 64: PacBio HiFi?

gymreklab: thanks, I changed to "PacBio HiFi" to clarify

heliziii (commented):
Hi Melissa, I went through your changes and all is good. Thank you.

heliziii (approved):

heliziii (approved):

PR #234, opened 2024-08-30
feat: Add options to facilitate debugging annotaTR runs on large files

gymreklab:
This PR introduces multiple options to annotaTR that help with debugging issues that arose when processing large files.

1. Added option `--region` to enable running annotaTR on a specified genomic region (chr:start-end). Helpful for debugging.
2. Added option `--update-ref-alt` to force annotaTR top copy over the ref/alt alleles from the reference panel. Helpful in cases where we have run `bcftools merge` in which case allele sequences may be modified. This in some cases causes problems since the INFO/END field is not updated accordingly, causing parsing HipSTR records to fail (related to the offsets set here: https://github.com/gymrek-lab/TRTools/blob/master/trtools/utils/tr_harmonizer.py#L340). Required adding function `CheckAlleleCompatibility` to annotaTR.
3. Added ability to output to bgzipped VCF output with `--outtype gzvcf`. Useful when output files are huge and we will end up zipping them anyway.
4. It adds option `--warn-on-AP-error` which results in skipping loci where checks on AP fields fail. In these cases, rather than the program quitting, we output nan values for dosages. In particular the checks this is relevant to are:

* Checking if the AP1/2 fields exist
* Checking if they sum to more than 1
* Checking for negative values
* Checking if normalized values end up being >=2.1 or <=-0.1

Most of these invalid AP cases should still never happen. We have encountered rare cases where values sum to more than 1, likely due to rounding errors in cases with huge numbers of alleles. The main motivation is in cases where we run annotaTR on huge VCF files which takes many hours only to encounter a bad AP field at the very end and crash, or when the vast majority of AP fields are fine but a few problematic loci cause the whole run to fail.

Other specific changes related to this option: (1) Added option `strict` to `GetDosages()`. This defaults to `true`, in which case we throw `ValueError` for the cases above. If this is `false`, we output a warning and return all dosage values as `np.nan`. (2) Regardless of whether the `strict` option is set, added info to the error/warning messages about which locus was problematic to help with tracking down those cases.

Finally, I added a couple additional tests unrelated to these when attempting to get good test coverage on `annotaTR.py`.

## Checklist

* [ x] I've checked to ensure there aren't already other open [pull requests](../../../pulls) for the same update/change
* [ x] At the top of the PR, I've [listed any open issues that this PR will resolve](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword). For example, "resolves #0" if this PR resolves issue #0
- [ x] I've explained my changes in a manner that will make it possible for both users and maintainers of TRTools to understand them
* [ x] All directories with large test files are listed in [the "exclude" section](https://python-poetry.org/docs/pyproject/#include-and-exclude) of our pyproject.toml so that they do not appear in our PyPI distribution. All new files are also smaller than 0.5 MB.
* [ x] I've updated the relevant REAMDEs with any new usage information and checked that the newly built documentation is formatted properly
* [x ] (if applicable) I've updated the pyproject.toml file with any changes I've made to TRTools's dependencies, and I've run `poetry lock --no-update` to ensure the lock file stays up to date and that our dependencies are locked to their minimum versions


trtools/utils/tr_harmonizer.py
```diff
@@ -1141,10 +1147,19 @@ def GetDosages(self,
 
             # Check AP field. allow wiggle room for rounding
             if np.any(np.sum(ap1, axis=1) > 1.1) or np.any(np.sum(ap2, axis=1) > 1.1):
-                raise ValueError("AP1 or AP2 field summing to more than 1 detected")
+                error_msg = "{}:{} AP1 or AP2 field summing to more than 1 detected".format(self.chrom, self.pos)
+                if strict:
+                    raise ValueError(error_msg)
+                else:
+                    common.WARNING(error_msg)
+                    return np.array([np.nan]*self.GetNumSamples())
```
aryarm, line 1155: At the moment, this will return NaN for all samples even if only one had an improper AP value. Would it be better to only return NaN for the few bad samples, instead?
```suggestion
            failing = (np.sum(ap1, axis=1) > 1.1) | (np.sum(ap2, axis=1) > 1.1)
            if np.any(failing):
                error_msg = "{}:{} AP1 or AP2 field summing to more than 1 detected".format(self.chrom, self.pos)
                if strict:
                    raise ValueError(error_msg)
                else:
                    common.WARNING(error_msg)
                    ap1[failing,:] = np.nan
                    ap2[failing,:] = np.nan
```

trtools/utils/tr_harmonizer.py
```diff
@@ -1141,10 +1147,19 @@ def GetDosages(self,
...
             if np.any(ap1 < 0) or np.any(ap2 < 0):
-                raise ValueError("Negative AP1 or AP2 fields detected")
-
+                error_msg = "{}:{} Negative AP1 or AP2 fields detected".format(self.chrom, self.pos)
+                if strict:
+                    raise ValueError("Negative AP1 or AP2 fields detected")
+                else:
+                    common.WARNING(error_msg)
+                    return np.array([np.nan]*self.GetNumSamples())
```
aryarm, line 1162: Same comment as above. Did you make an explicit decision to output NaN for all samples instead of just the ones that failed this check?
```suggestion
            failing = (ap1 < 0) | (ap2 < 0)
            if np.any(failing):
                error_msg = "{}:{} Negative AP1 or AP2 fields detected".format(self.chrom, self.pos)
                if strict:
                    raise ValueError("Negative AP1 or AP2 fields detected")
                else:
                    common.WARNING(error_msg)
                    ap1[failing,:] = np.nan
                    ap2[failing,:] = np.nan
```

trtools/utils/tr_harmonizer.py
```diff
@@ -1167,12 +1182,18 @@ def GetDosages(self,
...
                 # Normalize to be between 0 and 2
                 dosages = (unnorm_dosages-2*self.min_allele_length)/(self.max_allele_length-self.min_allele_length)
-                assert not (np.any(dosages>=2.1) or np.any(dosages<=-0.1))
+                if (np.any(dosages>=2.1) or np.any(dosages<=-0.1)):
+                    error_msg = "{}:{} Error normalizing dosages: value >=2.1 or <=-0.1 detected".format(self.chrom, self.pos)
+                    if strict:
+                        raise ValueError(error_msg)
+                    else:
+                        common.WARNING(error_msg)
+                        return np.array([np.nan]*self.GetNumSamples())
```
aryarm, line 1191: Same comment as above. If you couldn't compute dosages for some samples, could you still output dosages for the rest of them? Or would it be better to indicate complete uncertainty at this locus?
```suggestion
                failing = (dosages>=2.1) | (dosages<=-0.1)
                if np.any(failing):
                    error_msg = "{}:{} Error normalizing dosages: value >=2.1 or <=-0.1 detected".format(self.chrom, self.pos)
                    if strict:
                        raise ValueError(error_msg)
                    else:
                        common.WARNING(error_msg)
                        dosages[failing] = np.nan
```

trtools/annotaTR/tests/test_annotaTR.py
```diff
@@ -85,6 +88,25 @@ def test_OutTypes(args, vcfdir):
...
+    # Should get pgen error if input VCF has fewer
+    # TRs than ref panel
+    args.vcf = os.path.join(vcfdir, "beagle", "beagle_imputed_noTRs.vcf.gz")
+    args.ref_panel = os.path.join(vcfdir, "beagle", "beagle_tinyrefpanel.vcf.gz")
+    args.match_refpanel_on = "locid"
+    args.dosages = "bestguess_norm"
+    args.outtype = ["pgen"]
+    args.out = "test"
+    retcode = main(args)
+    assert retcode==1
```
aryarm, line 109: I think there might be some leftover files from running this test
```suggestion
    assert retcode==1
    for fname_ext in ("pgen", "pvar", "psam"):
        if os.path.exists("test."+fname_ext):
            os.remove("test."+fname_ext)
```

aryarm (commented):
It's interesting to see this get refined and closer and closer to production! I have just a few small comments which are scattered throughout. Thanks for asking me for a review!

One thing that wasn't immediately clear to me: How does `pgenlib.PgenWriter` handle `np.nan` values? Are those getting encoded properly as missing? And, if so, maybe it would be good to have a test that checks for that via `pgenlib.PgenReader` ? I can help write such a test, if you'd like.

gymreklab:
Thanks for these comments @aryarm!

* I updated the VCF output options as suggested (I wasn't totally sure if compressed BCF files should still be `.bcf` or should they be `.bcf.gz`? )

* Regarding setting all samples with bad AP fields to np.nan vs. only the failing samples: I had deliberately set them all to np.nan for now. This is because it happens very rarely (like 2 out of 50K loci which are actually removed from our cleaned up panel) and at what look to be problematic loci, and I would rather just exclude those entire loci from downstream analysis. I am also not sure what sort of biases get introduced if we remove only the failing samples. Would could always change in the future to only set the failing ones to np.nan but my preference is to keep it this way for now.

* For testing the downstream pgen output with np.nan values: curious on ideas for the best way to do this. For now, I have added a test file to force this case (`trtools/testsupport/sample_vcfs/beagle/beagle_badap.vcf.gz`) (used in `test/cmdline_tests.sh`). Running `plink2 --validate` on this file doesn't return any problems. Are there other checks we should do?

aryarm (approved):
ok, sounds good! I think it just uses `.bcf` (not `.bcf.gz`), so we should be fine

I also added commit 692a65bed421205eb3bf6cdb628c81ab955d841e to test that missing values are properly being written to the PGEN file output with `--warn-on-AP-error`

Issue #238, opened 2024-10-22
IUPAC handling in ExpansionHunter


raphaelbetschart:
Hi,

I'm running into some issues when using EnsembleTR with my ExpansionHunter VCF files.
The main issue is that the reference catalogue includes IUPAC nucleotides (for instance repeat structure `AARRG` at position chr4:39348424-39348479, taken from [https://github.com/Illumina/ExpansionHunter/blob/master/variant_catalog/hg38/variant_catalog.json](https://github.com/Illumina/ExpansionHunter/blob/master/variant_catalog/hg38/variant_catalog.json). This leads to an error in `utils.py` in the following line:
```
nucToNumber={"A":0,"C":1,"G":2,"T":3}
``` 

Should I just remove this entry or change the reference catalogue manually?
GangSTR and HipSTR catalogue both have the repeat structure `AAAAG`.

Thanks,
Raphael

aryarm:
Thanks for reporting this issue, @raphaelbetschart! We've made an attempt at fixing it in https://github.com/gymrek-lab/EnsembleTR/pull/32

Can you try the new code and let us know if you're still having issues by reopening this issue?

raphaelbetschart:
Hi @aryarm, the new code fixed the issue, thanks!
I found another small issue: `EnsembleTR --version` still reports version 1.0.0, but I can see that 1.0.1 was downloaded.

PR #239, opened 2024-10-30
fix: allow for IUPAC codes in `GetCanonicalMotif()`

aryarm:
partially addresses #238 

This will allow TRTools's `GetCanonicalMotif()` to work for sequences that have IUPAC codes. However, this will not entirely solve the problem described in #238, since any function that calls `GetCanonicalMotif()` will now have to handle the IUPAC codes somehow.

### DIFF
--- a/trtools/utils/tests/test_utils.py
+++ b/trtools/utils/tests/test_utils.py
@@ -115,6 +115,8 @@ def test_GetCanonicalMotif():
     assert(utils.GetCanonicalMotif("TTGTT")=="AAAAC")
     assert(utils.GetCanonicalMotif("")=="")
     assert(utils.GetCanonicalMotif("cag")=="AGC")
+    assert(utils.GetCanonicalMotif("AARRG")=="AARRG")
+    assert(utils.GetCanonicalMotif("YARRG")=="ARRGY")
 
 # GetCanonicalOneStrand
 def test_GetCanonicalOneStrand():
--- a/trtools/utils/utils.py
+++ b/trtools/utils/utils.py
@@ -14,7 +14,6 @@ import scipy.stats
 
 import trtools.utils.common as common # pragma: no cover
 
-nucToNumber={"A":0,"C":1,"G":2,"T":3}
 
 def LoadSingleReader(
         vcf_loc: str,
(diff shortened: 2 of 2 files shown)


trtools/utils/utils.py
```diff
@@ -387,9 +386,9 @@ def GetCanonicalMotif(repseq):
     repseq_r = GetCanonicalOneStrand(ReverseComplement(repseq))
```
LiterallyUniqueLogin, line 386: We also need to change ReverseComplement to properly reverse complement all of the other codes - for instance, the complement of whichever character stands for "A or G" is the character that stands for "T or C".

trtools/utils/tests/test_utils.py
```diff
@@ -115,6 +115,8 @@ def test_GetCanonicalMotif():
     assert(utils.GetCanonicalMotif("TTGTT")=="AAAAC")
     assert(utils.GetCanonicalMotif("")=="")
     assert(utils.GetCanonicalMotif("cag")=="AGC")
+    assert(utils.GetCanonicalMotif("AARRG")=="AARRG")
+    assert(utils.GetCanonicalMotif("YARRG")=="ARRGY")
```
LiterallyUniqueLogin, line 119: We should have a test here or in the reverse complement tests that confirms that when reverse complements are taken, the non-ACTG characters get handled appropriately.

LiterallyUniqueLogin (changes requested):
Aside from the reverse complement issue, whether or not this should be done depends on how it is used or intended to be used. I see no uses of GetCanonicalMotif or ReverseComplement in TRTools, so I cannot evaluate that and instead leave that choice to you.

aryarm:
Thanks for the suggestions! I tried my best to implement them in c1f78eeb73393a9642398911b9c585496b173af4

I think the use-case is to use `GetCanonicalMotif()` to discover whether variant records from different callers are mergeable in EnsembleTR:
https://github.com/gymrek-lab/EnsembleTR/blob/43104bfac4f98e5ea7500e2220500c21e6f337c5/ensembletr/vcfio.py#L230

aryarm:
great, thanks for the review, Jonathan!

PR #245, opened 2024-12-11
docs: add note about bcftools version for merging to annotaTR

### LINKED ISSUE #244: annotaTR - bcftools merge related issue for large cohorts
Hi,

First of all, thanks once again for developing TRTools! I have been using the newly implemented tool annotaTR particularly (v6.1.0), to annotate imputed TR VCFs with TRDS information to be used in plink2 for association testing (we have discussed this before, thank you also for the discussions on this earlier again!).

annotaTR now works really great and accurate based on the many tests I have performed, but the only problem I have currently is due to related to "bcftools merge" step actually.

If I use a large cohort (such as >40,000 samples), I try to split the input VCFs to be imputed into chunks of 8,000 samples for the Beagle imputation step, and as you recommend as well here (https://github.com/gymrek-lab/TRTools/blob/tr-gwas-tutorial/doc/VIGNETTE-GWAS-TUTORIAL.rst#step-14-merging-results-from-each-batch).

However, I realized that the chromosome-wide annotaTR fails and throws an error and exit when coming across with certain imputed TRs in the imputed VCFs (if I run --region specific annotaTR run, it works well, unless there is a problematic TR found in the provided region). The common pattern for these problematic imputed TRs are related to the order of different ALT alleles.

As you know bcftools merge trims the alleles (unfortunately), but this problem is solved by "--match-refpanel-on locid --update-ref-alt" options very nicely for most TRs; however still at certain TRs I see that the order of ALT alleles are impacted, and in these cases --update-ref-alt option cannot solve these cases. For instance, ALT1, ALT2, ALT3, ALT4 order in the original file becomes ALT1, ALT3, ALT2, ALT4 in the bcftools merge output, causing the problem.

The issue is perhaps expected as you also indicated it in the vignette (https://trtools.readthedocs.io/en/stable/source/annotaTR.html: "Use with caution as this assumes allele order is exactly the same between the refpanel and target VCF"), but I wanted to ask your advice on resolving this issue. For imputing large cohorts where you have to use bcftools merge, how do you avoid this problem, can you please help me with this? Thank you very much in advance!

Of note, I am using the most recent v4 imputation reference panel you have provided which have TR IDs (https://github.com/gymrek-lab/EnsembleTR/blob/fix-ref/README.md#version-iv-of-reference-snptr-haplotype-panel-for-imputation-of-tr-variants), and the most recent TRTools version (v6.1.0). Here is an example command I am using for annotaTR which works when --region parameter is used, but fails for chromosome-wide runs:

`annotaTR --vcf input_TRimputed_file_chr1.vcf.gz --out output_annotaTR_chr1 --vcftype hipstr --outtype pgen vcf --vcf-outtype z --chunk-size 10000 --dosages beagleap_norm --warn-on-AP-error --ref-panel ensembletr_refpanel_v4_chr1.vcf.gz --update-ref-alt`

And the error is as below:

>Traceback (most recent call last):
  File "/software/bin/annotaTR", line 10, in <module>
    sys.exit(run())
             ^^^^^
  File "/software/lib/python3.12/site-packages/trtools/annotaTR/annotaTR.py", line 721, in run
    retcode = main(args)
              ^^^^^^^^^^
  File "/software/lib/python3.12/site-packages/trtools/annotaTR/annotaTR.py", line 650, in main
    raise ValueError("--update-ref-alt set but the REF/ALT fields"
ValueError: --update-ref-alt set but the REF/ALT fields at chr1:2508612 are incompatible between the refpanel and target VCF

gymreklab:
This PR makes some minor clarifications to the annotaTR readme.

* Adds a note that older bcftools versions might switch allele order when doing bcftools merge, and that we tested on v1.20
* Minor clarification to options for `--match-refpanel-on`.

Resolves #244 

## Checklist

* [x ] I've checked to ensure there aren't already other open [pull requests](../../../pulls) for the same update/change
* [ x] At the top of the PR, I've [listed any open issues that this PR will resolve](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword). For example, "resolves #0" if this PR resolves issue #0
- [ x] I've explained my changes in a manner that will make it possible for both users and maintainers of TRTools to understand them
* [x ] All directories with large test files are listed in [the "exclude" section](https://python-poetry.org/docs/pyproject/#include-and-exclude) of our pyproject.toml so that they do not appear in our PyPI distribution. All new files are also smaller than 0.5 MB.
* [x ] I've updated the relevant REAMDEs with any new usage information and checked that the newly built documentation is formatted properly
* [ x] (if applicable) I've updated the pyproject.toml file with any changes I've made to TRTools's dependencies, and I've run `poetry lock --no-update` to ensure the lock file stays up to date and that our dependencies are locked to their minimum versions

### DIFF
--- a/trtools/annotaTR/README.rst
+++ b/trtools/annotaTR/README.rst
@@ -103,6 +103,7 @@ where:
 Additional relevant options:
 
 * :code:`--match-refpanel-on <string>`: indicates how to match loci between the reference panel and the target VCF. Options: locid, rawalleles, trimmedalleles (Default:locid)
+
     * **locid** matches on the ID in the VCF file. If your reference panel does not have informative IDs for TRs (e.g. all are set to "."), this option will not work and annotaTR will output an error
     * **rawalleles** means loci are matched on :code:`chrom:pos:ref:alt`. Note if you merged samples in your target VCF file using :code:`bcftools merge`, you should instead use the **trimmedalleles** option below, since bcftools will modify alleles to remove common sequence (see `this issue <https://github.com/samtools/bcftools/issues/726>`_)
     * **trimmedalleles** means loci are matched on :code:`chrom:pos:ref:alt` but ref and alt alleles are trimmed to remove common prefixes/suffixes.
--- a/trtools/annotaTR/README.rst
+++ b/trtools/annotaTR/README.rst
@@ -104,10 +104,10 @@ Additional relevant options:
 
 * :code:`--match-refpanel-on <string>`: indicates how to match loci between the reference panel and the target VCF. Options: locid, rawalleles, trimmedalleles (Default:locid)
     * **locid** matches on the ID in the VCF file. If your reference panel does not have informative IDs for TRs (e.g. all are set to "."), this option will not work and annotaTR will output an error
-    * **rawalleles** means loci are matched on :code:`chrom:pos:ref:alt`
-    * **trimmedalleles** means loci are matched on :code:`chrom:pos:ref:alt` but ref and alt alleles are trimmed to remove common prefixes/suffixes. The trimmedalleles option must be used if you merged samples in your target VCF file using :code:`bcftools merge`, since that tool will modify alleles to remove common sequence (see `this issue <https://github.com/samtools/bcftools/issues/726>`_)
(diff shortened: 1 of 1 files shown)


nicholema (approved):
The explanation looks good to me

PR #253, opened 2025-04-04
feat: output dosage and count in the .pvar file from annotaTR

yli091230:
## Checklist

* [x] I've checked to ensure there aren't already other open [pull requests](../../../pulls) for the same update/change
* [x] I've prefixed the title of my PR according to [the conventional commits specification](https://www.conventionalcommits.org/). If your PR fixes a bug, please prefix the PR with `fix: `. Otherwise, if it introduces a new feature, please prefix it with `feat: `. If it introduces a breaking change, please add an exclamation before the colon, like `feat!: `. If the scope of the PR changes because of a revision to it, please update the PR title, since the title will be used in our CHANGELOG.
* [x] I've explained my changes in a manner that will make it possible for both users and maintainers of TRTools to understand them
* [x] I've added tests for any new functionality. Or, if this PR fixes a bug, I've added test(s) that replicate it
* [x] All functions, modules, classes etc. still conform to [numpy docstring standards](https://numpydoc.readthedocs.io/en/latest/format.html)

## Description

Imputed TRs output from the annotaTR may be bi-allelic and contains rare alleles (only few count in a large cohort).  This PR makes annotaTR output a new INFO filed named `DSCOUNT` contains the non-zero dosage and their counts number in the `pvar` file. The script has been tested and an example output is provided below:
```bash
##fileformat=VCFv4.2
##INFO=<ID=DSLEN,Number=2,Type=Float,Description="Minimum and maximum dosages, used if normalization was applied">
##INFO=<ID=DSCOUNT,Number=1,Type=String,Description="list of dosage with non-zero count and their counts. use for filtering">
#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO
chr21	14280176	.	A	T	.	.	DSLEN=6.50,6.50	DSCOUNT=6.5;6
chr21	14286385	.	A	T	.	.	DSLEN=14.00,20.00	DSCOUNT=17.0,18.0,18.0,18.0;2,2,1,1
chr21	14289514	.	A	T	.	.	DSLEN=16.50,16.50	DSCOUNT=16.5;6
chr21	14291382	.	A	T	.	.	DSLEN=17.00,21.00	DSCOUNT=20.0;6
chr21	14291456	.	A	T	.	.	DSLEN=8.50,20.50	DSCOUNT=12.5,15.5,14.5;1,2,3
chr21	14292535	.	A	T	.	.	DSLEN=4.00,8.00	DSCOUNT=6.0,7.0;3,3
```

### DIFF
--- a/trtools/annotaTR/annotaTR.py
+++ b/trtools/annotaTR/annotaTR.py
@@ -256,6 +256,40 @@ def GetLocusKey(record, match_on=RefMatchTypes.locid):
     else:
         raise ValueError("Invalid match_refpanel_on=%s"%match_on)
 
+def GetAlleleCount(record):
+    """
+    Count the allele number that identified in samples
+
+    Parameters
+    ----------
+    record : cyvcf2.Variant
+        record to get get genotype and allele information of all samples
+
+    Returns:
+    ----------
+        allele_count_str: str
+            contains non zero count allele length and their counts
+            sperated by a ";", e.g. "1,2,3;200,169,300".
+    """
+    genotypes = record.genotypes
+    all_gt_list = np.array(genotypes)[:,:2]
+    unique_gt, counts = np.unique(all_gt_list.flatten(), return_counts=True)
+    if isinstance(record.ALT, list):
+        alt_list = record.ALT
+    else:
+        alt_list = [record.ALT]
+    all_alleles = np.array([record.REF] + alt_list)
+    repeat_len = record.INFO.get("PERIOD")
+    allele_repeat_n = np.char.str_len(all_alleles)/repeat_len
+    allele_repeat_n = np.round(allele_repeat_n, decimals=1)
+
+    unique_allele_str = ",".join(allele_repeat_n[unique_gt].astype(str))
+    counts_str = ",".join(counts.astype(str))
+
+    allele_count_str = f"{unique_allele_str};{counts_str}"
+
+    return allele_count_str
+
 def LoadMetadataFromRefPanel(refreader, vcftype, match_on=RefMatchTypes.locid,
         ignore_duplicates=False):
     """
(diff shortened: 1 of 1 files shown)


nicholema (commented):
The added feature looks good to me

aryarm:
one quick question: where do we usually use these info flags? would it make sense to have add a flag that turns this output off in case folks don't want it? or will it always be helpful to have?

yli091230:
> one quick question: where do we usually use these info flags? would it make sense to have a command line switch that turns this output off in case folks don't want it and they would prefer to save some space? or will it always be helpful to have?
Thanks for the comments. We noticed lots of imputed TRs are bi-allelic or with multi-alleles (usually 3 alleles) with major alleles with very high allele frequencies (e.g. >99.99%). Those loci will be problematic for GWAS, showing crazy effect size with great p-values. We want to use this DSCOUNT field to filter out those problematic alleles. It only write into the .pvar file. I kept as much as raw information there to enable flexible filtering options. I feel it worth to keep it there since it would take too much space.

trtools/annotaTR/annotaTR.py
```diff
@@ -256,6 +256,44 @@ def GetLocusKey(record, match_on=RefMatchTypes.locid):
     else:
         raise ValueError("Invalid match_refpanel_on=%s"%match_on)
 
+def GetAlleleCount(record):
```
aryarm, line 259:
```suggestion
def GetAlleleCount(record: cyvcf2.Variant):
```
How similar is this method to [`tr_harmonizer.GetAlleleCounts`](https://github.com/gymrek-lab/TRTools/blob/6bdc596db02dba1e38f7ab11fba8cb20377e0eaf/trtools/utils/tr_harmonizer.py#L1420)? Why do we need a new function for this? Maybe you can describe the differences between the two methods in the function signature of this one?

aryarm (commented):
> lots of imputed TRs dosages (or sum of TR lengths) are bi-allelic or with multi-alleles

ok, that makes sense! it sounds like we should always keep that in the .pvar output, in that case

It might be a good idea to add a test to the tests in [`test_annotaTR.py`](https://github.com/gymrek-lab/TRTools/blob/master/trtools/annotaTR/tests/test_annotaTR.py) to confirm that the .pvar output is as we would expect. Also, it might be good to have a separate test for the new `GetAlleleCount()` method

